Issue #9, opened 2017-02-12
Raise error upon construction of out-of-bounds datetime?


spencerkclark:
In pydata/xarray#1252 I am working on a ```NetCDFTimeIndex``` that is intended to bring some of the features of pandas's DatetimeIndex (for now namely field accessors and partial datetime string indexing) to time indexes that use ```netcdftime._netcdftime.datetime``` objects.  

Currently one can construct non-sensical datetimes using ```netcdftime._netcdftime.datetime``` objects:
```
In [1]: from netcdftime import DatetimeNoLeap

In [2]: DatetimeNoLeap(1, 45, 45)
Out[2]: netcdftime._netcdftime.DatetimeNoLeap(1, 45, 45, 0, 0, 0, 0, -1, 1)
```
Would it be possible for this kind of expression to raise an error?  This would be nice, because that way if one tries to index with a slice involving an out-of-bounds datetime, it would automatically cause an error, rather than behave in the fashion below:  
```
In [1]: import xarray as xr

In [2]: from xarray.conventions.netcdftimeindex import NetCDFTimeIndex

In [3]: from netcdftime import DatetimeNoLeap

In [4]: dates = [DatetimeNoLeap(1, 1, 1), DatetimeNoLeap(1, 2, 1), DatetimeNoLeap(2, 1, 1), DatetimeNoLeap(2, 2, 1)]

In [5]: da = xr.DataArray([1, 2, 3, 4], coords=[NetCDFTimeIndex(dates)], dims=['time'])

In [6]: da.sel(time=slice(DatetimeNoLeap(1, 1, 1), DatetimeNoLeap(1, 45, 45)))
Out[6]:
<xarray.DataArray (time: 2)>
array([1, 2])
Coordinates:
  * time     (time) object    1-01-01 00:00:00    1-02-01 00:00:00
```
cc @shoyer

jhamman:
This should absolutely be done. We should be able to provide similar checks to what is provided in the `datetime` module.

```Python-traceback
In [1]: import datetime

In [2]: datetime.datetime(2000, 1, 45, -4)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-2492d881556f> in <module>()
----> 1 datetime.datetime(2000, 1, 45, -4)

ValueError: day is out of range for month```

Issue #8, opened 2017-02-12
netcdftime.datetime refers to DatetimeProlepticGregorian


spencerkclark:
It would be nice if the super class for all netcdftime datetime objects (```netcdftime._netcdftime.datetime```) were exposed in the public API; this would allow one to succinctly do type checking.

Currently, ```netcdftime.datetime``` refers to the ```DatetimeProlepticGregorian``` object.  Is this intended, or could it be changed to point to ```netcdftime._datetime.datetime```?
```
In [1]: from netcdftime import datetime, DatetimeAllLeap

In [2]: datetime(1, 1, 1)
Out[2]: netcdftime._netcdftime.DatetimeProlepticGregorian(1, 1, 1, 0, 0, 0, 0, -1, 1)

In [3]: test = DatetimeAllLeap(1, 1, 1)

In [4]: isinstance(test, datetime)
Out[4]: False

In [5]: from netcdftime._netcdftime import datetime as super_datetime

In [6]: isinstance(test, super_datetime)
Out[6]: True
```
cc @shoyer
xref: https://github.com/pydata/xarray/pull/1252#discussion_r100679865

jhamman:
closed via #12

PR #12, opened 2017-03-28
Raise ValueError upon construction of invalid datetime object

spencerkclark:
Addresses #8 and #9

This PR switches ```netcdftime.datetime``` from pointing to ```DatetimeProlepticGregorian``` to the generic ```netcdftime._netcdftime.datetime``` object and adds logic to raise a ```ValueError``` if a constructed date is out of bounds for a given calendar.  

Locally all the tests pass when I make these changes, but I'm not sure if they have unintended consequences for netcdftime or netcdf4-python.  Please let me know if you have any comments or suggestions on my implementation.  In particular, this is my first time working with cython; hopefully what I have here is a decent start.  Many thanks.

### DIFF
--- a/netcdftime/__init__.py
+++ b/netcdftime/__init__.py
@@ -1,6 +1,6 @@
 from ._netcdftime import utime, JulianDayFromDate, DateFromJulianDay
 from ._netcdftime import _parse_date, date2index, time2index
-from ._netcdftime import DatetimeProlepticGregorian as datetime
+from ._netcdftime import datetime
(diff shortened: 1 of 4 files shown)


jhamman:
@spencerkclark - can you resolve the conflicts here?

spencerkclark:
@jhamman yes, I think things should be up to date now.

jhamman:
@jswhit - do you have any thoughts based on @spencerkclark's comments above?

jswhit:
Can't think of any reason this would cause problems.

PR #16, opened 2017-07-03
fix issue #685: wrong time zone conversion

mdecker:
Timezone calculation logic has been corrected.
Existing tests have been modified as they also used faulty logic.
Additionally, explicit tests for this problem have been added.


jswhit:
This has already been commited to netcdf4-python, so I'm going ahead and merging this.

Issue #17, opened 2017-08-02
timezone formats not supported correctly


mdecker:
The timezone parsing does not seem to follow the same standard that I can find on Wikipedia (https://en.wikipedia.org/w/index.php?title=ISO_8601&oldid=793312108#Time_zone_designators). Unfortunately, I cannot find any primary ISO source for the exact timezone format, so I have to rely solely on Wikipedia for the time being.
The page referenced in the code (https://github.com/Unidata/netcdftime/blob/1904de5c81fff7e40f3bbb23906dc2db83d7aaa1/netcdftime/_netcdftime.pyx#L33) is also not available any more. But judging from the archived version (https://web.archive.org/web/20161009150515/http://delete.me.uk/2005/03/iso8601.html), it suffers some - but not all of the format errors (see below).

What I think is wrong:
- Colons are optional in the timezone specification (according to Wikipedia). Right now the timezone string "+01" cannot be recognized, even though this as well as "+0100" should be valid.
- Both hour and minute specification should consist of exactly two numbers - not one OR two. This was actually done right in the JavaScript code quoted as reference. If this is fixed, then there won't be any ambiguities introduced by making the colon optional, either.

Both problems are relatively easy to fix by slightly modifying the regexes and maybe the parser code itself. If we can agree on the above points, I can work on a pull request to correct this.

mdecker:
some examples:

    >>> import netcdftime
    # correct: no timezone/UTC
    >>> netcdftime._parse_date("2017-05-01 00:00:00Z")
    (2017, 5, 1, 0, 0, 0, 0)
    # works, but should not due to bad format because hour and minute fields are only of len 1
    >>> netcdftime._parse_date("2017-05-01 00:00:00+1:0")
    (2017, 5, 1, 0, 0, 0, 60.0)
    # works as expected
    >>> netcdftime._parse_date("2017-05-01 00:00:00+01:00")
    (2017, 5, 1, 0, 0, 0, 60.0)
    # does not work
    >>> netcdftime._parse_date("2017-05-01 00:00:00+01")
    (2017, 5, 1, 0, 0, 0, 0)
    # also does not work
    >>> netcdftime._parse_date("2017-05-01 00:00:00+0100")
    (2017, 5, 1, 0, 0, 0, 0)

mdecker:
[...]

PR #18, opened 2017-08-18
fix for issue #17

mdecker:
Not as clean as I would like it to be because some ancient non-standard time strings have to be supported. Improvements welcome!


jswhit:
Seems reasonable to me.  @jhamman - if I don't hear from you by the end of the week I'll go ahead and merge. 

Since it's not clear when the transition to a stand-alone netcdftime package will be completed, I created a copy of this at https://github.com/Unidata/netcdf4-python/pull/704.

jhamman:
LGTM. Thanks!

mdecker:
TravisCI currently fails because _netcdftime.c was not regenerated. Can you do this or should I make a new PR?

PR #22, opened 2018-02-13
backport fixes for negative years (with tests) from netcdf4-python

jswhit:
with these all the tests in netcdf4-python pass


jhamman:
Thanks @jswhit!

PR #23, opened 2018-02-13
require cython and remove _netcdftime.c

jhamman:


jhamman:
@jswhit - this seems to work. Cython is listed in the requirements already so I think this is good to go. Am I missing anything?

Issue #21, opened 2018-02-13
Use Appveyor for CI on Windows


jhamman:
I think we should probably test this package on windows. I have no experience setting up Appveyor so I'm looking for someone that does and wants to see this package supported on Windows.

jhamman:
closing. @ocefpaf already did this and somehow I missed it.

jhamman:
So, I was kind of right. @ocefpaf did add an Appveyor config but something seems to be keeping it from running on master or on PRs.

ocefpaf:
I used my AppVeyor instance to test it when I sent the PR but I don't think anyone activated it for this repo.

PS: the link in the README, https://ci.appveyor.com/project/jswhit/netcdftime/branch/master, seems to be wrong, maybe @jswhit has the rights to activate AppVeyor here.

jhamman:
I think you are right. I do not have full admin rights here so @jswhit will have to link to Appveyor.

jswhit:
Done

jswhit:
appveyor tests fail, looks like the same issue I just fixed in netcdf4-python (https://github.com/Unidata/netcdf4-python/issues/758)

jswhit:
I fixed appveyor.yml in pull request #22

jhamman:
closed via #22.

PR #24, opened 2018-02-14
Zero-pad year in __str__

### LINKED ISSUE #19: Consider switching to a repr with a zero-padded year?
xref: https://github.com/pydata/xarray/pull/1252#discussion_r123914548

When encoding datetime-like objects for storage in netCDF files, xarray follows the standard convention (converting dates to a series of numbers representing some time unit since a reference date).  The units are encoded as a string attribute of a variable in a netCDF file, e.g. `'days since 2000-01-01 00:00:00'`.  In the process of constructing this string upon saving the file, [xarray uses the repr of the datetime object stored in the array](https://github.com/spencerkclark/xarray/blob/NetCDFTimeIndex/xarray/conventions/coding.py#L229).  Additionally, xarray's datetime decoding logic [depends on the use of the `pandas.Timestamp` constructor](https://github.com/spencerkclark/xarray/blob/NetCDFTimeIndex/xarray/conventions/coding.py#L150) to assess whether a reference date with a standard calendar (encoded as a string) can be represented using a `Timestamp` object (allowing the time series to be decoded fully without the optional dependency of netCDF4) or needs to be represented using a `netcdftime` object (requiring the optional dependency of netCDF4 to be decoded).

In experimenting with round-tripping arrays containing datetime objects outside the Timestamp-valid range to netCDF files and back to xarray objects within pydata/xarray#1252, the non-zero-padded repr of `netcdftime` objects (used in encoding the dates) and the datetime parser used by pandas in attempting to decode the dates do not always play well together.  See for example:
```
In [1]: import xarray as xr

In [2]: from netcdftime import DatetimeProlepticGregorian

In [3]: da = xr.DataArray([DatetimeProlepticGregorian(1, 1, 1), DatetimeProlepticGregorian(1, 2, 1)])

In [4]: da
Out[4]:
<xarray.DataArray (dim_0: 2)>
array([netcdftime._netcdftime.DatetimeProlepticGregorian(1, 1, 1, 0, 0, 0, 0, -1, 1),
       netcdftime._netcdftime.DatetimeProlepticGregorian(1, 2, 1, 0, 0, 0, 0, -1, 1)], dtype=object)
Dimensions without coordinates: dim_0

In [5]: da.to_dataset(name='time').to_netcdf('test-roundtrip.nc')

In [6]: xr.open_dataset('test-roundtrip.nc')
Out[6]:
<xarray.Dataset>
Dimensions:  (dim_0: 2)
Dimensions without coordinates: dim_0
Data variables:
    time     (dim_0) datetime64[ns] 2001-01-01 2001-02-01
```

Upon closer inspection, one can find that this could be fixed by using a repr with a zero-padded year when encoding the datetimes:
```
In [7]: ds = xr.open_dataset('test-roundtrip.nc', decode_times=False)

In [8]: ds.time
Out[8]:
<xarray.DataArray 'time' (dim_0: 2)>
array([ 0, 31])
Dimensions without coordinates: dim_0
Attributes:
    units:     days since    1-01-01 00:00:00
    calendar:  proleptic_gregorian

In [9]: ds.time.attrs['units'] = 'days since 0001-01-01 00:00:00'

In [10]: xr.decode_cf(ds)
/Users/spencerclark/xarray-dev/xarray/xarray/conventions/coding.py:416: RuntimeWarning: Unable to decode time axis into full numpy.datetime64 objects, continuing using dummy netCDF4.datetime objects instead, reason: dates out of range
  result = decode_cf_datetime(example_value, units, calendar)
Out[10]: /Users/spencerclark/xarray-dev/xarray/xarray/conventions/coding.py:435: RuntimeWarning: Unable to decode time axis into full numpy.datetime64 objects, continuing using dummy netCDF4.datetime objects instead, reason: dates out of range
  calendar=self.calendar)

<xarray.Dataset>
Dimensions:  (dim_0: 2)
Dimensions without coordinates: dim_0
Data variables:
    time     (dim_0) object 0001-01-01 0001-02-01
```
Would using repr with a zero-padded year possibly make sense for `netcdftime` or should we look for an alternative solution to this issue?

spencerkclark:
Fixes #19 

Previously `__str__` for netcdftime datetime objects used `strftime` with the format string `'%Y-%m-%d %H:%M:%S'` to convert datetime objects to strings.  This does not zero-pad the year:
```
In [1]: from netcdftime import DatetimeNoLeap

In [2]: str(DatetimeNoLeap(1, 1, 1))
Out[2]: '   1-01-01 00:00:00'
```

Note that this is not exactly ISO8601 format, but this PR ensures that the string representation of netcdftime datetime objects matches that for standard datetime objects for non-four-digit years.  E.g.:
```
In [1]: from datetime import datetime

In [2]: from netcdftime import DatetimeNoLeap

In [3]: str(datetime(1, 1, 1))
Out[3]: '0001-01-01 00:00:00'

In [4]: str(DatetimeNoLeap(1, 1, 1))
Out[4]: '0001-01-01 00:00:00'
```


spencerkclark:
@jhamman should Appveyor be triggered here?  I'm wondering if we may want to reopen #21 in light of https://github.com/Unidata/netcdftime/pull/15#issuecomment-307384019.

jhamman:
Yes, it should be. I'll reopen.

ocefpaf:
@spencerkclark until someone with rights here activates AppVeyor you can check it on my instance:

https://ci.appveyor.com/project/ocefpaf/netcdftime

It is failing now BTW but it is unrelated to this PR. I'll send a fix in a moment.

jhamman:
@spencerkclark - these changes look good to me. Can you sync with master then ping me when the tests pass again?

spencerkclark:
@jhamman great, checks are all green now.

PR #25, opened 2018-02-14
Fix "batch not found issue"

ocefpaf:
This should fix `AppVeyor` builds. I also added `numpy 1.14` to the matrix.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,12 +11,22 @@ environment:
       CONDA_INSTALL_LOCN: C:\\Miniconda-x64
 
     - TARGET_ARCH: x64
+      CONDA_NPY: 114
+      CONDA_PY: 27
+      CONDA_INSTALL_LOCN: C:\\Miniconda-x64
+
+    - TARGET_ARCH: x64
       CONDA_NPY: 112
       CONDA_PY: 36
       CONDA_INSTALL_LOCN: C:\\Miniconda35-x64
 
     - TARGET_ARCH: x64
       CONDA_NPY: 113
+      CONDA_PY: 36
+      CONDA_INSTALL_LOCN: C:\\Miniconda35-x64
+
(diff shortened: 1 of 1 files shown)


jhamman:
@ocefpaf - can you sync with master after #22.

appveyor.yml
```diff
@@ -36,20 +46,19 @@ install:
 
     # Add path, activate `conda` and update conda.
     - cmd: call %CONDA_INSTALL_LOCN%\Scripts\activate.bat
-    - cmd: conda config --set always_yes yes --set changeps1 no --set show_channel_urls true
-    - cmd: conda update conda
-    - cmd: conda config --add channels conda-forge --force
-    - cmd: conda config --system --add pinned_packages defaults::conda
+    - cmd: conda.exe config --set always_yes yes --set changeps1 no --set show_channel_urls true
+    - cmd: conda.exe update conda
+    - cmd: conda.exe config --add channels conda-forge --force
```
ocefpaf: Calling via `conda.exe` is "safer" and guaranteed to work.

ocefpaf: This removes the `defaults` pin that @jswhit added in https://github.com/Unidata/netcdftime/commit/42ccc0566aba5016cc957d1f04c06b7d8611a40f and keep using `conda-forge`.

I don't have any strong opinion on this for `netcdftime` but I'd love to revert this in netcdf4-python (PR soon) b/c conda-forge usually has newer `hdf5` and `netcdf-c` and they are compiled with extra flags/options, so it is better to test against that.

jswhit: +1  - please do submit a pull request for netcdf4-python.  I really don't know what I'm doing with appveyor - so I'd much rather have your solution.

jhamman:
Thanks @ocefpaf. Merging now.

PR #30, opened 2018-02-21
fix namespace conflict for date2index

### LINKED ISSUE #26: Duplicate def of `date2index`
There are two separate definitions of `date2index` in _netcdftime.pyx, at lines 286 and and 1309. I assume one of these should go away.

jhamman:
closes #26


jswhit:
Looks good.

PR #38, opened 2018-03-21
Rename module from netcdftime to cftime

### LINKED ISSUE #36: netcdftime cannot be used with netCDF4-python<=1.3.1
As discussed in #32, when netCDF4 is installed its version of netcdftime overrides any installation of this package. It is very difficult to expect scientific python users to have the most up to date version of netCDF4, so up to date that it does not exist yet. This gives very strange behavior that is difficult to debug unless you've already run into it.

There is a really simple fix for this, which is to rename the netcdftime package, or provide an alias package name. This would remove a massive package incompatibility problem.

At the very least, netcdftime should require that these older versions of netCDF4 are *not* installed. This would allow me to tell users that they need to install netcdftime to get X functionality, and have a guarantee that when they do so they will not run into these import problems. One way to do this would be to check in netCDF4 whether netCDF4 is installed, and if it is installed to add a dependency on netCDF4 having at least the required version.

mcgibbon:
I replaced every occurence of 'netcdftime' with 'cftime'. This will break tests until github/travis/appveyor names are updated. Resolves #36 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ before_install:
 
 # Test source distribution.
 install:
-  - python setup.py sdist && version=$(python setup.py --version) && pushd dist  && pip install netcdftime-${version}.tar.gz && popd
+  - python setup.py sdist && version=$(python setup.py --version) && pushd dist  && pip install cftime-${version}.tar.gz && popd
 
 script:
   - if [[ $BUILD_DOCS == "true" ]]; then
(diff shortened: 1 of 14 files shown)


jhamman:
Should we merge this and then rapidly iterate on the CI / Pypi / etc. changes?

jhamman:
@jswhit - can you rename the repository?

jhamman:
hmmm, doesn't seem to be working for me. I can access the settings page but updating the package name fails.

jswhit:
Doesn't work for me either

jswhit:
Maybe a unidata admin has to do it. I think @dopplershift created the netcdftime project.

dopplershift:
Done.

PR #39, opened 2018-03-28
Add only_use_cftime_datetimes flag to num2date

### LINKED ISSUE #37: Add a version of num2date that strictly returns netcdftime.datetime objects?
In response to https://github.com/pydata/xarray/pull/1252#issuecomment-374711420, @shoyer wrote:
> I think netcdftime should have a version of num2date that *always* returns
netcdftime.datetime objects. It's really error prone to functions that turn
different types dependent on input values.

Would this be something folks would be open to?  Somewhat selfishly this would make things easier downstream in pydata/xarray#1252, because we would only need to keep track of two datetime types for indexing and it would hopefully simplify the logic to encode datetimes for faithful roundtripping via `netcdftime.num2date`.

`num2date` largely does this already.  I think it currently only returns `datetime.datetime` objects in the following situations (though please correct me if there are others):
- If dates are encoded using a calendar type of `'gregorian'` or `'proleptic_gregorian'`
- If the units argument provided to `num2date` includes a date after the beginning of the Gregorian calendar and the dates are encoded using a calendar type of `'standard'`.

If we could agree on how this alternative version of `num2date` would work in these situations I would be happy to put together a PR.  I'm not sure if we might want to define a new function or expose this via a keyword argument in the existing `num2date`.

spencerkclark:
Closes #37 

I tried to go for the path of least resistance.  I added a flag to `DateFromJulianDay` that allows one to bypass the logic that uses `datetime.datetime` objects in certain situations (and exposed that flag in `netcdftime.num2date`).  If `only_use_netcdftime_datetimes` is `True` we have the following behavior:
- For `calendar='standard'`, `DatetimeGregorian` objects are always used
- For `calendar='gregorian'`, `DatetimeGregorian` objects are always used
- For `calendar='proleptic_gregorian'`, `DatetimeProlepticGregorian` objects are always used

If `only_use_netcdftime_datetimes` is `False` (which is the default), then the original behavior occurs.  

Does this seem like something reasonable?  I'm happy to discuss things further.

### DIFF
--- a/netcdftime/_netcdftime.pyx
+++ b/netcdftime/_netcdftime.pyx
@@ -178,7 +178,7 @@ def date2num(dates,units,calendar='standard'):
             return cdftime.date2num(dates)
 
 
-def num2date(times,units,calendar='standard'):
+def num2date(times,units,calendar='standard',only_use_netcdftime_datetimes=False):
     """num2date(times,units,calendar='standard')
 
     Return datetime objects given numeric time values. The units
@@ -200,6 +200,10 @@ def num2date(times,units,calendar='standard'):
     Valid calendars `'standard', 'gregorian', 'proleptic_gregorian'
     'noleap', '365_day', '360_day', 'julian', 'all_leap', '366_day'`.
     Default is `'standard'`, which is a mixed Julian/Gregorian calendar.
+
+    **`only_use_netcdftime_datetimes`**: if False (default), datetime.datetime
+    objects are returned from num2date where possible; if True dates which
+    subclass netcdftime.datetime are returned for all calendars.
 
     returns a datetime instance, or an array of datetime instances with
     approximately millisecond accuracy.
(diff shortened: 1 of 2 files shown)


netcdftime/_netcdftime.pyx
```diff
@@ -669,15 +678,21 @@ def DateFromJulianDay(JD, calendar='standard'):
     if calendar in 'proleptic_gregorian':
```
spencerkclark: In looking at this logic, I noticed that as written this branch of the if-statement is used for both `calendar='proleptic_gregorian'` and `calendar='gregorian'`.  Therefore, while the next branch is `elif calendar in ('standard', 'gregorian')` it will never be used in the case of `calendar='gregorian'`.  Was this intended?

In other words, should this line be `if calendar == 'proleptic_gregorian'`?

spencerkclark:
Appveyor seems to be failing in setting up the environment for Python 2.7, so I suspect it is unrelated to the changes I've made here (the same thing seems to be happening in #38).

spencerkclark:
This is now updated in accordance with the renaming of the package to `cftime`; the option in `num2date` is now called `only_use_cftime_datetimes`.

jhamman:
@spencerkclark - Thanks, this looks good. I'd like @jswhit to have a quick look before merging but generally, its nice to have the added consistency!

PR #40, opened 2018-04-12
pin conda

ocefpaf:
This is a temporary fix for the AppVeyor failure. I'll work with appveyor to solve this so we can use latest conda there.


ocefpaf:
Actually this did not fixed all the items in the matrix :-/
See https://ci.appveyor.com/project/ocefpaf/netcdftime/build/1.0.23

I'll work on a long term solution soon.

Issue #11, opened 2017-03-28
Datetime comparison


jswhit:
copy of https://github.com/Unidata/netcdf4-python/issues/639

When a netcdftime datetime object is unable to be compared to another object, the __richcmp__ method raises a TypeError. In previous versions of the module the method would return NotImplemented, causing the appropriate method on the other object to be called and allowing classes to be written which could be compared against netcdftime datetime objects. With the TypeError behaviour this is no longer the case.

Can the previous behaviour be restored?

jswhit:
Seems like simply changing TypeError to NotImplementedError in __richcmp__ would fix this.  Are they any downsides to this?

shoyer:
CC @djkirkham

> Seems like simply changing TypeError to NotImplementedError in richcmp would fix this. Are they any downsides to this?

It needs to be `return NotImplemented` to properly defer to the implemention on the other object, not `raise NotImplementedError`

pdearnshaw:
This problem has arisen in a project I am working on so I thought to try this and got unexpected results, basically `cftime.datetime` objects become comparable to anything! ie. breaks `not_comparable_4` test in `test_richcmp`.

Is there anything else that can be done?

PR #53, opened 2018-07-09
Fix rich comparisons for datetime objects

pdearnshaw:
As identified in #11, rich comparisons for `cftime.datetime` objects with other unknown objects do not work as cftime raises a `TypeError` instead of returning `NotImplemented`. This pull request is to fix this.

While this works in Python 3, it should be noted that Python 2 is more problematic. The unit test `not_comparable_4` in `test_richcmp` will fail as `cftime.datetime` objects are treated as comparable with integer zero and so the `TypeError` is not raised when it is expected.

### DIFF
--- a/cftime/_cftime.pyx
+++ b/cftime/_cftime.pyx
@@ -1616,6 +1616,13 @@ Gregorial calendar.
                 raise TypeError("cannot compare {0!r} and {1!r} (different calendars)".format(self, other))
             return PyObject_RichCompare(dt.to_tuple(), to_tuple(other), op)
         else:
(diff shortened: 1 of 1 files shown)


jswhit:
The python 2.7 Travis tests are failing.  We can't merge until this is fixed.

jswhit:
I'm OK with this. @jhamman?

jhamman:
I'm also fine with this but would appreciate a test to validate the expected behavior.

jswhit:
I can merge this as soon as a test is added

jswhit:
I'm going to go ahead and merge this, without a test for now.

Issue #59, opened 2018-08-14
different requirements in conda recipe and setup.py


xylar:
`cftime` seems to think it requires a different set of packages to run in the conda recipe:
```
    - python
    - setuptools
    - numpy x.x
```
and in `setup.py` (via `requirements.txt`):
```
numpy
cython
setuptools>=18.0
```
The requirement for `cython` appears to be retained in such a way that a conda environment that includes `cftime` but not `cython` and tries to use make use of `pkg_resources` gets the error:
```
...
pkg_resources.DistributionNotFound: The 'cython' distribution was not found and is required by cftime
```
The error is resolved by installing `cython` or including it as an (unnecessary) dependency of any conda package that depends on `cftime` but it would be more convenient if `cython` were only included in `setup.py` in `setup_requires` and not in `install_requires`.

So far, I haven't been able to create a super simple test that reproduces the issue.  I see it when I try to build a conda package from a repo that I'm just starting to develop.  Here is where I seem to need to include `cython`, though I don't think it should be needed:
https://github.com/xylar/misomip1analysis/blob/initial_stub/conda/recipe/meta.yaml#L28

ocefpaf:
@xylar the conda-recipe here is only for Windows testing and I'm inclined to remove that altogether and use a bare bones AppVeoyr config. BTW, that is an issue with the recipe here and not `cftime`. (Due to its use of `python setup.py install` rather than `pip`.)

TL;DR you should use `cftime` from `conda-forge` if you need a conda package for that and not this recipe.

xylar:
@ocefpaf, thank you very much for the reply.

I think I am, indeed, using `cftime` from `conda-forge`.  As with the recipe that you are removing here, `cython` is not included as a dependency (which I think is correct).  The problem seems to remain that `pkg_resources` is somehow picking up the list of dependencies passed to `setup.py` rather than the one that is included in the `conda-forge` recipe, meaning that it raises this erroneous `DistributionNotFound` exception.

Have you seen this kind of thing anywhere in the past?

@czender and colleagues saw exactly the same error at least once before.  I'm not sure how they managed to resolve it:
https://github.com/jaws/jaws/issues/115

ocefpaf:
> I think I am, indeed, using cftime from conda-forge

OK. Let's move this discussion to the feedstock then. I'll check that in a moment...

xylar:
[...]

PR #61, opened 2018-08-14
No conda build

ocefpaf:
This reduces the amount of boilerplate code needed to run the tests on AppVeyor and avoid confusions like #59


ocefpaf:
@jswhit we should probably do the same on `netcdf4-python`. Let me know if you are interested and I'll send a PR.

(`conda-build` can hinder some failures and was not really designed for this kind of testing.)

jswhit:
yes, that would be great- thanks @ocefpaf

PR #65, opened 2018-08-16
fix for 'day out of range error' (issue #49)

jswhit:
Fix was to restore the fix added in https://github.com/Unidata/netcdf4-python/issues/75, which was removed at some point because we thought it was no longer needed.  Apparently it is.

Also removed import of python `math` module, used `numpy` instead.


jswhit:
Turns out the error is associated with numpy's 'even rounding' which causes 0.5 to be rounded to 0.0.  Using 'round half up' fixes this without having to add a small offset to the julian day.

PR #69, opened 2018-09-29
add month support to 360_day calendar

### LINKED ISSUE #68: decode 'months since' for 360_day calendars
The [IRI Data Library](https://iridl.ldeo.columbia.edu) contains tons of datasets with the following time attributes:
```
	float32 T(T) ;
		T:standard_name = time ;
		T:pointwidth = 1.0 ;
		T:long_name = Time ;
		T:calendar = 360 ;
		T:expires = 1538524800 ;
		T:gridtype = 0 ;
		T:units = months since 1960-01-01 ;
```
(edited to correct typo in original post)

([example dataset](https://iridl.ldeo.columbia.edu/SOURCES/.NOAA/.NCEP-NCAR/.CDAS-1/.MONTHLY/.Diagnostic/.surface/.temp/dods))

I would like to be able to open and decode these datasets in xarray, with time decoding handled by cftime.

There are two problems:
1. The calendar is not a valid CF calendar. It should be `360_day`. But that is easy to fix by rewriting the calendar attribute.
1. Even with `calendar = 360_day`, `months` is not considered a valid time unit

However, in a 360-day calendar month==30 days, so this should be valid.

This was discussed over in https://github.com/Unidata/netcdf4-python/issues/434#issuecomment-133083060, where @jswhit commented:
> A pull request allowing months since when calendar is 360_day would be welcome

rabernat:
fixes #68 

I only implemented `num2date` and not `date2num`.

Is it important to be able to go both ways?

### DIFF
--- a/cftime/_cftime.pyx
+++ b/cftime/_cftime.pyx
@@ -22,6 +22,7 @@ sec_units =      ['second', 'seconds', 'sec', 'secs', 's']
 min_units =      ['minute', 'minutes', 'min', 'mins']
 hr_units =       ['hour', 'hours', 'hr', 'hrs', 'h']
 day_units =      ['day', 'days', 'd']
+month_units =    ['month', 'months'] # only valid for 360_day calendar
 _units = microsec_units+millisec_units+sec_units+min_units+hr_units+day_units
 _calendars = ['standard', 'gregorian', 'proleptic_gregorian',
               'noleap', 'julian', 'all_leap', '365_day', '366_day', '360_day']
@@ -54,7 +55,7 @@ class real_datetime(datetime_python):
 gregorian = real_datetime(1582,10,15)
 
 def _datesplit(timestr):
-    """split a time string into two components, units and the remainder 
+    """split a time string into two components, units and the remainder
     after 'since'
     """
     try:
@@ -93,11 +94,11 @@ def _round_half_up(x):
     # 'round half up' so 0.5 rounded to 1 (instead of 0 as in numpy.round)
     return np.ceil(np.floor(2.*x)/2.)
 
-cdef _parse_date_and_units(timestr):
+cdef _parse_date_and_units(timestr, calendar):
     """parse a string of the form time-units since yyyy-mm-dd hh:mm:ss
     return a tuple (units,utc_offset, datetimeinstance)"""
     (units, isostring) = _datesplit(timestr)
-    if units not in _units:
+    if not ((units in month_units and calendar=='360_day') or units in _units):
         raise ValueError(
             "units must be one of 'seconds', 'minutes', 'hours' or 'days' (or singular version of these), got '%s'" % units)
     # parse the date string.
@@ -1083,7 +1084,7 @@ units to datetime objects.
         else:
             raise ValueError(
                 "calendar must be one of %s, got '%s'" % (str(_calendars), calendar))
-        units, tzoffset, self.origin = _parse_date_and_units(unit_string)
+        units, tzoffset, self.origin = _parse_date_and_units(unit_string, calendar)
         # real-world calendars limited to positive reference years.
         if self.calendar in ['julian', 'standard', 'gregorian', 'proleptic_gregorian']:
             if self.origin.year == 0:
@@ -1244,6 +1245,9 @@ units to datetime objects.
             jdelta = time_value / 24. - self.tzoffset / 1440.
         elif self.units in day_units:
             jdelta = time_value - self.tzoffset / 1440.
+        elif self.units in month_units:
+            # only possible for 360_day calendar
+            jdelta = time_value * 30. - self.tzoffset / 1440.
         else:
             raise ValueError('unsupported time units')
         jd = self._jd0 + jdelta
@@ -2175,4 +2179,3 @@ cdef tuple add_timedelta_360_day(datetime dt, delta):
     month = (month - 1) % 12 + 1
 
     return (year, month, day, hour, minute, second, microsecond, -1, 1)
-
--- a/test/test_cftime.py
+++ b/test/test_cftime.py
@@ -475,6 +475,10 @@ class cftimeTestCase(unittest.TestCase):
         d = num2date(0, units, calendar='360_day')
         self.assertEqual(d, Datetime360Day(0,1,1))
 
+        # issue XXX
+        d = num2date(1, 'months since 0000-01-01 00:00:00', calendar='360_day')
+        self.assertEqual(d, Datetime360Day(0,2,1))
+
         # issue 685: wrong time zone conversion
         # 'The following times all refer to the same moment: "18:30Z", "22:30+04", "1130-0700", and "15:00-03:30'
         # (https://en.wikipedia.org/w/index.php?title=ISO_8601&oldid=787811367#Time_offsets_from_UTC)


jswhit:
Thanks for the contribution!  I really think we should have it in both num2date and date2num.

davidhassell:
Sorry for the late intervention, but I think that this is at odds with the CF conventions.

CF says that a month is 365.242198781/12 days (http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#time-coordinate), but it does redefine it for any of the allowed calendars.

For this to work, the CF conventions would have to say explicitly that "a month is never a UDUNITS month but is always a *calendar year* over 12", which it does not, at this time.

I fear that this special, non-CF exception will be damaging to those of us who rely on the standard, as a dataset creator's intention can not be assumed to be known to all users of that dataset.

Thanks, David

rabernat:
CF Conventions do not explicitly define month lengths. That is done by udunits. Here is what CF Conventions say.

> We recommend that the unit year be used with caution. The Udunits package defines a year to be exactly 365.242198781 days (the interval between 2 successive passages of the sun through vernal equinox). It is not a calendar year. Udunits includes the following definitions for years: a common_year is 365 days, a leap_year is 366 days, a Julian_year is 365.25 days, and a Gregorian_year is 365.2425 days.
>
> For similar reasons the unit month, which is defined in udunits.dat to be exactly year/12, should also be used with caution.

So it seems that the process for changing this should go through udunits. Perhaps we should open an issue there?

> I fear that this special, non-CF exception will be damaging to those of us who rely on the standard, as a dataset creator's intention can not be assumed to be known to all users of that dataset.

I appreciate the need to respect the standard and go through proper channel. But do you really, honestly thing there are people out there providing datasets with a `360_day` calendar who want month units to be interpreted as 365.242198781/12?

rabernat:
Also worth noting that the current behavior of cftime is to raise an error with month units, rather than to interpret it as 365.242198781/12. So it would seem that the "standard" is already not being followed.

davidhassell:
The problem is less with UDUNITS, as they do not, and have never claimed to, support non-real-world calendars. I do not think that anyone *intentionally* uses the udunits `month` to mean anything other than a calendar month, but one cannot know this for sure, and that is why generic software for supporting CF really ought to follow the standard.

I would favour a behaviour whereby when a `months since ...` is encountered that its converted to `days since ...` with the conversion factor

davidhassell:
The problem is less with UDUNITS, as they do not, and have never claimed to, support non-real-world calendars. I do not think that anyone intentionally uses the udunits month to mean anything other than a calendar month, but one cannot know this for sure, and that is why generic software for supporting CF really ought to follow the standard.

I would favour a behaviour whereby when a `months since ...` is encountered that its converted to `days since ...` under the hood, unless a keyword like `thirty_day_months=True` to `num2date` is used.

The default `num2date` would then do something like (this code snippet is using an older netCDF4-python, but I hope that the idea is clear)
```
       if units in ('month', 'months'):
            # Convert months to days
            unit_string = unit_string.replace(units, 'days', 1)
            time_value = numpy.array(time_value)*365.242198781/12
        elif units in ('year', 'years', 'yr'):
            # Convert years to days
            unit_string = unit_string.replace(units, 'days', 1)
            time_value = numpy.array(time_value)*365.242198781

        u = netCDF4.netcdftime.utime(unit_string, self.calendar)        

        return u.num2date(time_value)
```
with the addition of the code in the pull request https://github.com/Unidata/cftime/pull/69/commits/49fc234781c04d81e79761af7919021c7a064085, that would be controlled by a keyword parameter in the call.

Then the standard is being followed,  the user would be alerted if something is awry (by virtue of getting very weird dates), and the user would have the option of overriding the default behaviour with a well defined, but non-standard, alternative.

This approach could also be extended to cover months in other well-defined calendars, like `366_day`, but that is a bit more fiddly! Extending to cover years on other calendars would be somewhat easier.

Thanks, David

jswhit:
How about a kwarg `days_in_month`, which is N/12 by default for calendars with a fixed number of days per year (`360_day`, `no_leap` and `all_leap`),  and is undefined for the other calendars (up to the user to specify)?  SImilarly, 'years since` could be supported with the same approach (using a `days_in_year` kwarg).

jswhit:
I think it's worth pointing out that a `months since` unit really only makes sense in the `360_day` calendar, since that calendar is defined to be made up of 12 30 day months per year.  So - I'm OK with implementing this for the `360_day` calendar first, since there is an obvious use case there, while continuing the discussion about what to do for other calendars.

lesserwhirls:
By calling out the definition of `year` and `month` explicitly, and providing a warning along the lines of "these do not mean what you would likely expect" (that language has been in since CF 1.0), I think CF is pretty clear in how one should go about interpreting the length of a month. So, I would say it is not following CF to convert UDUNIT month to 30 UDUNIT days for this calendar. It makes 100% sense to treat it as such in this case, but does not follow the spec. If `cftime` is supposed to strictly follow CF, then case closed? Not to say that alone is completely useful...

> How about a kwarg `days_in_month`, which is N/12 by default for calendars with a fixed number of days per year (`360_day`, `no_leap` and `all_leap`), and is undefined for the other calendars (up to the user to specify)? SImilarly, `years since` could be supported with the same approach (using a `days_in_year` kwarg).

This approach seems reasonable to me, but certainly not following CF. I think it's perhaps worth bringing up these "calendars with fixed number of days" cases on the CF list, possibly adding to the spec, but even then you'd need to decide how to handle data already written (pre-CF 1.8(?), if added?).

As a comparison, the CDM tries to be a bit more flexible with time, in which time as defined by CF is only part of what it supports (see [here](https://www.unidata.ucar.edu/software/thredds/current/netcdf-java/CDM/CalendarDateTime.html) for more info). In addition to the UDUNITS grammar of time (`period SINCE reference_date`), the CDM also supports what it calls the calendar_field unit grammar (`CALENDAR period SINCE reference_date`). So, for example, you can have:

`calendar months since 1990-03-01`

which makes clear one is talking about a calendar month. Same thing for year (i.e. `calendar years since 1990`). Not CF, but it's not the intent of the CDM to follow CF alone. Given John Caron's deep involvement with CF, this additional flexibility to the CDM makes me think that, again, the intent of CF on how month and year is to be interpreted is clear (currently).

My 2 cents (worth even less than that, really)...if you intend to follow CF alone, then stick to the spec, and try to extend the spec through the proper channels to accommodate the case of uniform years. If you intended to be more flexible, document the heck out of it, shout it from the rooftops, come up with a plan as to what to do if/when your assumptions break something that follows CF, and possibly rename the package to something other than `cftime`.

@ethanrd?

jswhit:
I just read the CF Time coordinates section again, and I still don't see where it says anything specific about the length of a month.  All I see is this:

```
We recommend that the unit year be used with caution. The Udunits package defines a year to be exactly 365.242198781 days (the interval between 2 successive passages of the sun through vernal equinox). It is not a calendar year. Udunits includes the following definitions for years: a common_year is 365 days, a leap_year is 366 days, a Julian_year is 365.25 days, and a Gregorian_year is 365.2425 days.

For similar reasons the unit month, which is defined in udunits.dat to be exactly year/12, should also be used with caution.
```

So it talks about udunits, but nowhere does it say that udunits == CF.  There are certainly many calendars that CF describes that udunits doesn't. Is CF a superset of udunits?

rabernat:
👍  to what @jswhit said, which is also what I said above (https://github.com/Unidata/cftime/pull/69#issuecomment-427070396). Is the udunits library part of the CF spec or not?

dopplershift:
The right thing to do, IMO, is raise this on the CF mailing list. The real question is what's being done right now in real-world data files, and are the changes proposed here going to result in incorrect behavior.

lesserwhirls:
In section 1.3, it says:

> The values of the units attributes are character strings that are recognized by UNIDATA’s Udunits package [UDUNITS] , (with exceptions allowed as discussed in Section 3.1, "Units" ).

My question would be what does "recognized by" mean? A character-by-character match (purely a unit namespace issue) or the definition of the unit as well? 

At the very least, it sounds like the right thing to do would be to go through CF to add an explicit exception, especially since CF provides a mechanism for making exceptions.

ethanrd:
CF section [3.1 Units](http://cfconventions.org/cf-conventions/cf-conventions.html#units) does not list the `calendar` attribute as an exception to the rule that "the value of the units attribute is a string that can be recognized by UNIDATA’s Udunits package" but I think it should.

From the first two sentences of CF section [4.4.1 Calendar](http://cfconventions.org/cf-conventions/cf-conventions.html#calendar):

> "In  order to calculate a new date and time given a base date, base time and a time increment one must know what calendar to use. For this purpose we recommend that the calendar be specified by the attribute calendar which is assigned to the time coordinate variable."

I believe the intent is that the calendar specified in a CF `calendar` attribute should be applied to all components of a time variable including the time increment. So, when `calendar = 360_day` and `units = months since ...`, a month should be considered 30 days as specified in the definition of "360_day" calendar.

ethanrd:
Yes, I agree with @dopplershift and @lesserwhirls that the right thing is to discuss this on the CF list and, probably, develop some language to make some calendars and exception to the UDUnits rule.

davidhassell:
Interpreting the units in light of the calendar could get messy. What is 12 months after 2004-02-29 in the Gregorian calendar? or one month after 2017-01-29? I think that there is no definitively right answer. Similarly, a month length of N/12 for fixed-length years wouldn't work well since, in general,  this would give non-integer month lengths (for N=365 or N=366).

What is being done with real world files right now depends on the software being used to read them. Library software that  follows the CF specification might well do something different to bespoke code written by someone with prior knowledge of the dataset creator's intent. Neither approach is wrong, although I would argue that the dataset creator should have known better than to use such units ... but that ship has clearly sailed!

I think that CF does a pretty good job of accepting that time _deltas_ in units of months or years are often  ill-defined, and also highlighting that the inherited (from UDUNITS) units of month and year are not what you hope they might be. Clarifying the relationship of "calendar" to UDUNITS would be no bad thing, though, as @ethanrd suggests.

rabernat:
Would someone else like to initiate the mailing list discussion, or should I do it?

And are we talking about [this](http://cfconventions.org/discussion.html) mailing list: cf-metadata@cgd.ucar.edu?

jswhit:
I would be in favor of deviating from UDUNITS and treating 'months since' as calendar months in the specific situation where the calendar specifies the length of the month as a fixed number of days (i.e. the 360-day calendar).  In such a situation, this is clearly what the user would expect.  Similarly, when the calendar specifies that a year has a fixed number of days, I think it's reasonable to assume that 'years since' refers to calendar years.  In all other cases, I think we should keep the current behavior (raise an error that says 'months since' and 'years since' are not supported).  I can start a discussion on the cf-metadata list if no one has already.

rabernat:
@jswhit - I support your view (obviously, since it's my PR)! I feel that this suggestion would have more weight on the cf-metadata list coming from you, since you understand the lingo and the history much better than I do. Thanks for offering to start the discussion.

jswhit:
http://mailman.cgd.ucar.edu/pipermail/cf-metadata/2018/020603.html

rabernat:
Arguably, not providing date2num would prevent people from creating such
problematic encodings. 😉

jswhit:
I created a new pull request #75 that includes both the date2num and num2date implementations, and is up to date with master.  I have followed the discussion on cf-metadata with interest, and now better appreciate all the pitfalls of calendar-based time units.  However, I still think that for the particular case of the 360_day calendar there is no ambiguity, and it would be useful to have this capability even if it does not strictly conform to the CF conventions (since there are apparently many datasets in the wild that could use this).

jswhit:
merged pull request #75

PR #76, opened 2018-10-27
change float128 to longdouble to fix compilation using MS compiler

jswhit:


sebastic:
It's not just the MS compiler that has issues with cftime 1.0.2, the Debian package build also failed on [i386](https://buildd.debian.org/status/fetch.php?pkg=cftime&arch=i386&ver=1.0.2-1&stamp=1540647017&raw=0), [armel](https://buildd.debian.org/status/fetch.php?pkg=cftime&arch=armel&ver=1.0.2-1&stamp=1540647362&raw=0) & [mips](https://buildd.debian.org/status/fetch.php?pkg=cftime&arch=mips&ver=1.0.2-1&stamp=1540647809&raw=0).

These two changes fix the build failure on i386 at least, probably also on the other architectures.

PR #81, opened 2018-11-11
Fix cftime.datetime.__richcmp__

bjlittle:
This PR re-addresses the now closed `cftime` issue https://github.com/Unidata/cftime/issues/11 (linked to legacy `netcdf4-python` issue https://github.com/Unidata/netcdf4-python/issues/639) and associated `cftime` PR https://github.com/Unidata/cftime/pull/53.

The current implementation of `cftime.datetime.__richcmp__` does not align with the old sought after behaviour alluded to in https://github.com/Unidata/cftime/issues/11, and from my perspective that's a bit of a problem... as it's now **not possible** to perform (previously supported and operational) comparison operations between `cftime.datetime` and other custom datetime objects i.e. more specifically an iris [PartialDateTime](https://github.com/SciTools/iris/blob/v2.2.0/lib/iris/time.py#L29) calendar agnostic objects.

For example, it was possible to do the following comparison using a `netcdftime.datetime` object `dt`,

```python
pdt1 <= dt < pdt2
```
where `pdt1` and `pdt2` are iris `PartialDateTime` objects.

As it stands at the moment with `cftime` 1.0.2.1, users are forced to do the equivalent comparison as follows, due to the implementation restrictions of `cftime.datetime.__richcmp__`,

```python
(pdt1 <= dt) and (pdt2 > dt)
```

as the current implementation of the `cftime.datetime.__richcmp__` method will raise a `TypeError` if a `cftime.datetime` object is a LHS operand participating in a rich comparison operation with another custom object (in Python2).

Apologies for the long winded introduction :cold_sweat: (I figured that context for this PR would be helpful!)... but this PR bridges the gap back to the old behaviour, and maintains closer consistency with the behaviour between Python2 and Python3 for `cftime.datetime`.

In a nutshell, for Python3 it's business as usual for `cftime.datetime.__richcmp__`. However for Python2, rich comparisons with other custom objects will be performed by delegating the comparison to the other object **only if** it advertises that it supports Cython `__richcmp__` **or** a suitable traditional rich comparison operator i.e. `__lt__`, `__le__`, `__eq__`, `__ne__`, `__gt__`, or `__ge__`, given the context of the current comparison.

A `TypeError` will **only** be raised otherwise, thus ensuring that the other object `__cmp__` behaviour is circumvented, and more importantly, the default Python object ID comparison is avoided. 

For example, if we are doing a Python2 `dt < pdt2` comparison, the `dt` object will **first check** that the other custom `pdt2` object supports either `__richcmp__` or `__gt__` operations, before delegating the comparison operation to the `pdt2` object via returning `NotImplemented`. Otherwise, a `TypeError` exception is raised.

Needless to say, the changes in this PR extends `cftime.datetime.__richcmp__` support such that it again permits users to do `pdt1` <= dt < pdt2`, whoop!

### DIFF
--- a/cftime/_cftime.pyx
+++ b/cftime/_cftime.pyx
@@ -2,7 +2,8 @@
 Performs conversions of netCDF time coordinate data to/from datetime objects.
 """
 
-from cpython.object cimport PyObject_RichCompare
+from cpython.object cimport (PyObject_RichCompare, Py_LT, Py_LE, Py_EQ,
+                             Py_NE, Py_GT, Py_GE)
 import cython
 import numpy as np
 import re
@@ -37,6 +38,9 @@ cdef int[12] _dpm_360  = [30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30]
 # Same as above, but SUM of previous months (no leap years).
 cdef int[13] _spm_365day = [0, 31, 59, 90, 120, 151, 181, 212, 243, 273, 304, 334, 365]
 cdef int[13] _spm_366day = [0, 31, 60, 91, 121, 152, 182, 213, 244, 274, 305, 335, 366]
+# Reverse operator lookup for datetime.__richcmp__
+_rop_lookup = {Py_LT: '__gt__', Py_LE: '__ge__', Py_EQ: '__eq__',
+               Py_GT: '__lt__', Py_GE: '__le__', Py_NE: '__ne__'}
 
 __version__ = '1.0.2.1'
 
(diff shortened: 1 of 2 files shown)


bjlittle:
Heads up @pdearnshaw ...

jswhit:
OK then - merging now.  Thanks for another useful contribution @bjlittle!

bjlittle:
@jswhit Thanks for responding so swiftly.

When are you aiming to schedule a tag and release? Just curious...

PR #88, opened 2018-12-02
add new badges

bjlittle:
This PR simply adds three new badges to the `README.md` to show the latest tag, the latest release, and the number of commits since the latest release - which is handy for users and developers, alike.


bjlittle:
@jswhit You can see an example of the additional badges [here](https://github.com/bjlittle/cftime/tree/add-commits-badge#cftime) on my associated development branch

PR #89, opened 2018-12-02
add pypi trove classifiers

bjlittle:
This PR adds PyPI Trove Classifers for `cftime`.


bjlittle:
@jswhit Not entirely sure what the `License` entry should be given the options of [PyPI Trove Classifiers](https://pypi.org/pypi?%3Aaction=list_classifiers) for `License`... any suggestions?

jswhit:
I think we can use `License :: OSI Approved`

PR #90, opened 2018-12-02
pytest alias

bjlittle:
This PR adds the necessary glue to allow a developer to use `python setup.py test` to run the tests using `pytest`.

I've also renamed the `test` directory to be `tests` (hope that's not too controversial :stuck_out_tongue_winking_eye:). As a result, this will require a minor change to the `recipe/meta.yaml` for the `conda-forge` recipe at some point.


jswhit:
Does this require pytest to be installed in order to run setup.py at all, or only 'setup.py test'?

bjlittle:
@jswhit Good question... I'll check that tomorrow for you

bjlittle:
@jswhit I'm gonna drop this PR,  it's just not necessary and I don't want to increase the minimum dependencies of `cftime`, thanks

PR #95, opened 2018-12-02
add bogus nanosecond attribute to real_datetime (issue #77)

### LINKED ISSUE #77: cftime.real_datetime not pandas compatible
See https://github.com/pandas-dev/pandas/issues/23419 for context.

The issue is following:

```python
import pandas, numpy, netCDF4
time=netCDF4.num2date([0,1,2,3,4,5,500,1000], 'days since 1801-10-01')
pandas.Series(numpy.arange(len(time)), index=time)
```

Fails with

```
AttributeError: 'real_datetime' object has no attribute 'nanosecond'
```

@spencerkclark says that this is due to the fact that netcdf4 now returns cftime objects per default (always). I guess it is a necessary change (having different time objects to deal with depending on the calendar is a pain), but it makes me wonder about the level of compatibility between cftime objects and the rest of the pydata stack.

I'm no specialist at all here, so I don't know what it would involve, but it would be nice if cftime objects could be handled by pandas.

The easiest fix seems to simply add a `nanoseconds` attribute to the `cftime.real_datetime` objects.

jswhit:
Closes #77

### DIFF
--- a/cftime/_cftime.pyx
+++ b/cftime/_cftime.pyx
@@ -65,6 +65,7 @@ class real_datetime(datetime_python):
     @property
     def dayofyr(self):
         return self.timetuple().tm_yday
+    nanonsecond = 0 # workaround for pandas bug (cftime issue #77)
 
 # start of the gregorian calendar
 gregorian = real_datetime(1582,10,15)


bjlittle:
@jswhit 👍

PR #98, opened 2018-12-03
remove cythonize so that cython not imported before setup_requires used

jswhit:
should fix issue #34, which crept back in via pull request #83

Issue #34, opened 2018-03-02
setup.py imports from Cython before checking for Cython


jsolbrig:
As discussed in Unidata/netcdf4-python#767, setup.py calls `from Cython.Build import cythonize` before it requires Cython to be installed.  As a consequence, building without Cython previously installed fails with the following error:

```
Traceback (most recent call last):
  File "setup.py", line 3, in <module>
    from Cython.Build import cythonize
ImportError: No module named Cython.Build
```

jswhit:
I think [this](pull request #35) should fix it (untested).  Based on https://stackoverflow.com/questions/37471313/setup-requires-with-cython

jswhit:
@jsolbrig, can you test this and let us know if it fixes the problem?

jsolbrig:
@jswhit Yep, that solved it.  I'm also able to build netcdf4-python against it.

jhamman:
closed via #35

terokinnunen:
Looks like this issue came back in release 1.0.3.
```
[terokinnunen@localhost tmp]$ python3 -m venv venv
[terokinnunen@localhost tmp]$ source venv/bin/activate
(venv) [terokinnunen@localhost tmp]$ pip install -U pip setuptools
Collecting pip
  Using cached https://files.pythonhosted.org/packages/c2/d7/90f34cb0d83a6c5631cf71dfe64cc1054598c843a92b400e55675cc2ac37/pip-18.1-py2.py3-none-any.whl
Collecting setuptools
  Using cached https://files.pythonhosted.org/packages/e7/16/da8cb8046149d50940c6110310983abb359bbb8cbc3539e6bef95c29428a/setuptools-40.6.2-py2.py3-none-any.whl
Installing collected packages: pip, setuptools
  Found existing installation: pip 18.0
    Uninstalling pip-18.0:
      Successfully uninstalled pip-18.0
  Found existing installation: setuptools 40.4.3
    Uninstalling setuptools-40.4.3:
      Successfully uninstalled setuptools-40.4.3
Successfully installed pip-18.1 setuptools-40.6.2
(venv) [terokinnunen@localhost tmp]$ pip install cftime
Collecting cftime
  Using cached https://files.pythonhosted.org/packages/80/23/cd5513106c709f1faa3623bb34a5284b55ea7dad4229103b68c154d8e39f/cftime-1.0.3.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
[...]

PR #101, opened 2018-12-05
re-fix cythonize

bjlittle:
Addresses #34 (again, sorry)

Closes #98

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,8 +3,13 @@ from __future__ import print_function
 import os
 import sys
 
-from Cython.Build import cythonize
 from setuptools import Command, Extension, setup
+
+# https://github.com/Unidata/cftime/issues/34
+try:
+    from Cython.Build import cythonize
+except ImportError:
+    cythonize = False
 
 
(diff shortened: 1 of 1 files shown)


bjlittle:
The coverage is back to ~87%, which is what we expect... and the `setup.py` will only `cythonize` if `Cython` is available (and it is for the travis-ci test environment, hence the coverage works correctly), otherwise `setup.py` won't `cythonize` the `Extension`, as per previously.

jswhit:
Hopefully it still will cythonize the extension even if cython is not installed, since cython is in setup_requires.  (that was the source of the problem reported in issue #34)   What is cythonize doing that distutils Extension doesn't to enable the coverage stuff to work?  Is it not using the `compiler_directives` and `language_level` kwargs?

jswhit:
Agreed.  I'll see if I can test in an environment without cython installed, just to make sure it works as expected.

jswhit:
Works as expect in a environment without cython, merging now....

bjlittle:
Thanks @jswhit and apologies for the carnage...

PR #108, opened 2018-12-15
No longer support replacing dayofwk and dayofyr

### LINKED ISSUE #106: dayofyr and dayofwk after cftime.datetime.replace
#66 enabled working `dayofyr` and `dayofwk` attributes for all cftime dates constructed manually, which is great!

I've noticed that if one constructs datetime objects using `replace`, that the `dayofyr` and `dayofwk` attributes are not automatically updated:
```
In [21]: date = cftime.DatetimeNoLeap(1, 2, 1)

In [22]: date.dayofyr
Out[22]: 32

In [23]: date.replace(year=2, month=5).dayofyr
Out[23]: 32
```
A workaround is to specify `dayofwk=-1` within `replace`: 
```
In [24]: date.replace(year=2, month=5, dayofwk=-1).dayofyr
Out[24]: 121
```
Is this intentional?  Should we not pass down the old `dayofyr` and `dayofwk` attributes to the new date object in `replace`?  I'm happy to provide a PR to change this behavior, if desired.

spencerkclark:
Closes #106

Thanks @jswhit -- this is sort of what I had in mind.  Here I simply no longer propagate the old `dayofyr` and `dayofwk` attributes from the original date to the new one.  I'm not sure if there was a use-case for replacing the `dayofyr` and `dayofwk` attributes of a date, since they are intrinsic properties of a given datetime.

I added a couple tests to this implementation too.  I'm happy to make any changes if you have any comments.

### DIFF
--- a/cftime/_cftime.pyx
+++ b/cftime/_cftime.pyx
@@ -1220,9 +1220,7 @@ Gregorial calendar.
                 "hour": self.hour,
                 "minute": self.minute,
                 "second": self.second,
-                "microsecond": self.microsecond,
-                "dayofwk": self.dayofwk,
-                "dayofyr": self.dayofyr}
+                "microsecond": self.microsecond}
 
         for name, value in kwargs.items():
             args[name] = value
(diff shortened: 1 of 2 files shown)


jswhit:
I like you version better - removing the other pull request.

spencerkclark:
Thanks!  This should be ready to merge once the CI finishes if you have no further comments.

PR #110, opened 2019-02-08
Feature tzaware

barronh:
Internally converting dates with timezones to naive UTC so that dates with timezones can be passed to JulianDayFromDate and date2num. Implicitly adds support to utime.date2num.

Functionality and tests have been added. Including python2.7 capabilities.


jswhit:
Looks good, merging now.  Thanks @barronh !

Issue #118, opened 2019-09-06
Show microseconds in str() output, and suggestied change to repr() output


davidhassell:
Hello,

Currently, fractions of seconds are not shown in the str() representation of a `cftime` date-time object:

```
In [5]: import cftime                                                                                                                                                             

In [6]: cftime.__version__                                                                                                                                                        
Out[6]: '1.0.3.4'

In [7]: dt = cftime.DatetimeGregorian(2000, 2, 13, 12, 34, 23, 123456)                                                                                                            

In [8]: repr(dt)                                                                                                                                                                  
Out[8]: 'cftime.DatetimeGregorian(2000, 2, 13, 12, 34, 23, 123456, 6, 44)'

In [9]: str(dt)                                                                                                                                                                   
Out[9]: '2000-02-13 12:34:23'
```

I would like to suggest that they are shown, in a ISOI 8601-compliant fashion (https://en.wikipedia.org/wiki/ISO_8601#General_principles), so that the str() output would look like:

```
>>> str(dt)
'2000-02-13 12:34:23.123456'
```

I am happy to provide a pull request if this idea is accepted.

On a related note, might it be possible to alter the repr() output to improve its readaibility? I fully appreciate that this is subjective, and the developers have the last word, but I'll describe how it could perhaps be changed in case you agree!

We currently have:
```
In [8]: repr(dt)                                                                                                                                                                  
Out[8]: 'cftime.DatetimeGregorian(2000, 2, 13, 12, 34, 23, 123456, 6, 44)'
```

I would like to suggest:

```
 >>> rep(dt)
'cftime.DatetimeGregorian(2000-02-13 12:34:23.123456)'
```

My use case for this is that I frequently inspect `numpy` arrays of `cftime` objects, and the current repr() output delivered by `numpy` is hard for a (this!) human to parse quickly. We currently have:

```
In [10]: dt2 = cftime.DatetimeGregorian(2000, 2, 14, 12, 34, 23, 123456)                                                                                                          

In [11]: dt3 = cftime.DatetimeGregorian(2000, 3, 13, 12, 34, 23)                                                                                                          

In [12]: import numpy                                                                                                                                                             

In [13]: numpy.array([dt, dt2, dt3])                                                                                                                                              
Out[13]: 
array([cftime.DatetimeGregorian(2000, 2, 13, 12, 34, 23, 123456, 6, 44),
       cftime.DatetimeGregorian(2000, 2, 14, 12, 34, 23, 123456, 0, 45),
[...]

PR #120, opened 2019-09-10
changes suggested by issue #118

davidhassell:
Changes to __repr__ and __str__ suggested by issue #118


jswhit:
repr tests need to be updated

Issue #126, opened 2019-10-22
date2num error when converting a DatetimeProlepticGregorian object in cftime 1.0.4


spencerkclark:
In version 1.0.3.4 of cftime the following succeeds without an error:
```
In [1]: import cftime

In [2]: d = cftime.DatetimeProlepticGregorian(1, 1, 1)

In [3]: cftime.date2num(d, 'days since 0001-01-01', 'proleptic_gregorian')
Out[3]: 0.0
```
However, with 1.0.4, the following occurs:
```
In [3]: cftime.date2num(d, 'days since 0001-01-01', 'proleptic_gregorian')
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-5b0bfcafe058> in <module>
----> 1 cftime.date2num(d, 'days since 0001-01-01', 'proleptic_gregorian')

cftime/_cftime.pyx in cftime._cftime.date2num()

AttributeError: 'cftime._cftime.DatetimeProlepticGregorian' object has no attribute 'tzinfo'
```
We came across this in the xarray test suite, xref: https://github.com/pydata/xarray/issues/3427.  Was this an intentional change?

jswhit:
Seems to be related to 

https://github.com/Unidata/cftime/pull/110

@barronh, could you take a look?

jswhit:
Looks like a simple fix for a typo that slipped through in PR #110 (PR #127).  Assuming the xarray test suite runs cleanly with this fix, I'll push a 1.0.4.1 release.

barronh:
I’m at a conference thru tomorrow, so I haven’t been able to look at it. I trust that you’re right. I could look on Thursday.

jswhit:
Will go ahead and merge since the fix is trivial.

jswhit:
F

PR #127, opened 2019-10-22
fix for issue #126

jswhit:


spencerkclark:
Many thanks for looking into this so quickly @jswhit -- I can confirm that the xarray test suite passes after https://github.com/pydata/xarray/pull/3430 and this update to cftime.  

Going forward we will test against the master version of cftime, https://github.com/pydata/xarray/pull/3431, so hopefully we can catch minor things like this before a release is made.  Sorry for the trouble this time around.

PR #138, opened 2019-12-17
Added daysinmonth attribute to DateTime classes

huard:
Datetime objects now have a daysinmonth attribute returning the number of days in the month, similar to what pandas does. 

Not extensively tested...


jswhit:
Looks good - thanks @huard !

Issue #134, opened 2019-11-20
Error in num2date cinversion


rkouznetsov:
Originally posted to https://github.com/Unidata/netcdf4-python/issues/981, but is still reproducible with freshly-cloned cftime.

```
  #!/usr/bin/env python
 import numpy as np
 import cftime as nc4
 iarr=np.arange(86400, dtype=np.int32)
 units='hours since 2018-01-01 00:00:00 UTC'
 dates1 = nc4.num2date(iarr,units)[-5:]
 dates2 = nc4.num2date(iarr[-5:],units)
 print dates1 == dates2
```
 Results in
` [ True False True True False]`
 Should be
` [ True True True True True]`

 Environment:
 Ubuntu 18.04, python-numpy 1:1.13.3-2ubuntu1, cftime (revision ea6882365f1077baf3d920f790a0ad7d29874394).
 
 Thank you!

jswhit:
Microseconds are now included in the comparison (https://github.com/Unidata/cftime/issues/118) - and the 2nd and 5th dates are off by 13 microseconds. The calculation itself has an accuracy of about 100 milliseconds.  Perhaps we should modify the comparison to ignore differences of less than a specified tolerance?  @davidhassell - what do you think?

rkouznetsov:
Hi, Thank you for prompt response. I do not think that the comparison is an issue. I can imagine a case when 13 microseconds difference matters, so comparison did a good job.   I would say, that in given example the issue is in handling dates as float/double. The integer number of seconds from midnight should never result in non-zero microseconds. In unix the internal format for time (struct  timeval) is two integers: seconds and microseconds. It would be great to have something like that in cftime...

jswhit:
cftime time uses julian day/fractional julian day for it's arithmetic - so it's subject to usual issues of floating point precision/comparison.  Probably a better way to compare dates given a specified tolerance would be something like this:

```python
import numpy as np
import cftime as nc4
iarr=np.arange(86400, dtype=np.int32)
units='hours since 2018-01-01 00:00:00 UTC'
dates1 = nc4.num2date(iarr,units)[-5:]
dates2 = nc4.num2date(iarr[-5:],units)
datediff = [diff.microseconds for diff in dates1-dates2]
print(np.abs(datediff) < 100) # tolerance of 100 microseconds
```

rkouznetsov:
[...]

PR #139, opened 2019-12-30
add use_only_python_datetimes kwarg to num2date

jswhit:
issue #134


rkouznetsov:
Thank you! The modifications seem to do the job. A couple of concerns:

1. The request of python_datetime requires the right setting both optional arguments explicitly.
In case one forgets  to set only_use_cftime_datetimes=False, the argument only_use_python_datetimes is ignored, which might be a bit couterintuitive. If this is going to be merged "as is", I would consider including  a wrapper (e.g. num2pydate) to get it right without getting too much into the internal logic of the sub. 

2. I wonder if there is _any_ use case when user wants the sub to _guess_ which object to return. If you decide do kill a guessing part (which I would strongly support), a single key would be perfectly sufficient.

Makes sense?

jswhit:
I have added the `num2pydate` convenience function.  Regarding having the function guess what type of object to return - in general this is a bad idea.  Users want to know what type of objects they are going to get ahead of time.

Issue #148, opened 2020-03-17
Error installing with pip in Python3.8


lschwetlick:
Trying to install cftime in a clean Python 3.8.0 environment on MacOs (using `pip3 install cftime --no-cache-dir`, the "--no-cache-dir" option in pip was just to make sure I got a fresh copy) I get the following error:

```
Collecting cftime
  Downloading https://files.pythonhosted.org/packages/05/ec/9233d3b5ba54f9fa50782945527926b987eee38aa9d792133a6713c2a759/cftime-1.1.1.tar.gz (48kB)
     |████████████████████████████████| 51kB 451kB/s 
    ERROR: Command errored out with exit status 1:
     command: /Users/lisa/Documents/virtual_envs/sw_arviz6/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/gk/07t7dq7n01v45sj7k8295dqc0000gn/T/pip-install-gvw0ka37/cftime/setup.py'"'"'; __file__='"'"'/private/var/folders/gk/07t7dq7n01v45sj7k8295dqc0000gn/T/pip-install-gvw0ka37/cftime/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base pip-egg-info
         cwd: /private/var/folders/gk/07t7dq7n01v45sj7k8295dqc0000gn/T/pip-install-gvw0ka37/cftime/
    Complete output (5 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/var/folders/gk/07t7dq7n01v45sj7k8295dqc0000gn/T/pip-install-gvw0ka37/cftime/setup.py", line 5, in <module>
        import numpy
    ModuleNotFoundError: No module named 'numpy'
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

The error does not occur:
- using python<3.8
- When I explicitely install numpy first and then cftime

PR #149, opened 2020-03-17
follow pep517

ocefpaf:
Closes https://github.com/Unidata/cftime/issues/148

~~Don't merge this just yet. I want to refine it a little bit.~~

Actually any further changes would be better in another PR to avoid confusion. This is good to go.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools>=41.2", "cython", "wheel", "numpy"]
+build-backend = "setuptools.build_meta"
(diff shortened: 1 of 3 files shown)


ocefpaf:
@jswhit at the moment `setup.py` covers both options `cythonize = False/True`. We can limit that to always `True` if we want with this PR b/c cython will be guaranteed to be present at install time.

jswhit:
Looks good - thanks @ocefpaf !

lschwetlick:
Great, that was super fast, thanks!

Issue #152, opened 2020-03-19
AttributeError: 'cftime._cftime.DatetimeGregorian' object has no attribute 'isoformat'


zssherman:
Hello,

Our CI started to fail recently with the error:
AttributeError: 'cftime._cftime.DatetimeGregorian' object has no attribute 'isoformat'

I was going through the cftime change log and had a hard time pinpointing this, I might be missing it, but was wondering if there was a recent change to the DateTime Gregorian class that would be causing this error? 

Here is our code that fails:

        time_dim = ('string_length', )
        units = radar.time['units']
        start_dt = cftime.num2date(radar.time['data'][0], units)
        if start_dt.microsecond != 0:
            # truncate to nearest second
            start_dt -= datetime.timedelta(microseconds=start_dt.microsecond)
        end_dt = cftime.num2date(radar.time['data'][-1], units)
        if end_dt.microsecond != 0:
            # round up to next second
            end_dt += (datetime.timedelta(seconds=1) -
                       datetime.timedelta(microseconds=end_dt.microsecond))
      ->  start_dic = {'data': np.array(start_dt.isoformat() + 'Z', dtype='S'),
                     'long_name': 'UTC time of first ray in the file',
                     'units': 'unitless'}
       AttributeError: 'cftime._cftime.DatetimeGregorian' object has no attribute 'isoformat'

Versions are:
Python 3.7
netcdf 1.5.3
cftime 1.1.1

Thanks for your time and I can provide more info in needed,
Zach S.

jswhit:
There never was an `isoformat` cftime.datetime method.  However, until recently the num2date function returned python datetime instances wherever possible, and python datetime does have an isoformat method.  If you wish to recover the previous default behaviour in num2date, you can use the keyword `only_use_cftime_datetimes=False`.

zssherman:
Ah gotcha makes sense. I'll give the keyword a try, much appreciated!

dopplershift:
Would there be any interest in a PR to add `isoformat()`? It would make duck type usage of the `cftime` objects along side the standard library `datetime.datetime` in many contexts (like my own Siphon library) much easier.

jswhit:
Sure that would be great @dopplershift

jswhit:
Should be able to extract the code from the __str__ method, put it into isformat and have __str__ return self.isoformat()

jswhit:
@dopplershift - how about [this](https://github.com/Unidata/cftime/pull/168)?

[...]

PR #168, opened 2020-04-23
add isformat method (issue #152)

jswhit:


jswhit:
Should the default separator be ' ' for backwards compatibility, or 'T' for compatibility with python datetime?

dopplershift:
Awesome, thanks for taking it on!

Backward compatibility with...?  I'm confused since this method didn't exist here. Wikipedia's page on [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) indicates that you can't use a space as a separator. (Under Combined Date and Time Representations).

jswhit:
The `__str__` method now uses `isoformat`.  printing an instance now uses a blank for a separator.  I suppose I can keep that, but make the default for isoformat 'T' as in python datetime.

dopplershift:
That seems like a reasonable way to structure it. I hadn't caught the change to `__str__`.

PR #174, opened 2020-05-20
Return the default values of dayofwk and dayofyr when calendar is '' (issue #173)

### LINKED ISSUE #173: Consequences of changing the default calendar in cftime.datetime
If I set a non-gregorian calendar in the initialization of a `cftime.datetime` object, then it is ignored in the `daysinmonth` calculation:
```
In [1]: import cftime

In [2]: cftime.__version__
Out[2]: '1.1.3'

In [3]: d = cftime.datetime(2000, 3, 3, calendar='360_day')

In [4]: d.calendar
Out[4]: '360_day'

In [5]: d.daysinmonth
Out[5]: 31
```

This is surprising! The `dayofwk` attribute does indeed take the calendar into account.

On a related note, is it still possible to encode a datetime with an unspecified calendar. This was possible when the default calendar was `''`, and was extremely useful for encoding a datetime that is meant to be interpreted in the context of one or calendars that are not yet known. At inspection time, the calendar value of `''` tells my code to define a new datetime in the appropriate calendar so that comparisons can be made. I can't set `calendar=''`, as other things break (as expected):
```
In [31]: e = cftime.datetime(2000, 3, 4, calendar='')

In [32]: e.timetuple()

ValueError: unsupported calendar
```
Is `cftime.datetime` any different, now, to `cftime.DatetimeGregorian`? If so, can the calendar default on the  base class `cftime.dateime`  be reverted to `''`. In this case, `dayofwk`, etc. would be `None` (say) when called on the base class. Is there a use case for having the default as `'standard'`?

Apologies for realizing these questions when these changes were being discussed - I have just found out when my units tests failed with 1.1.3! 
Thanks, David

davidhassell:
Fixes #173

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,7 +1,7 @@
 version <> (release tag v<>rel)
 =====================================
-* Return -1 from dayofwk and dayofyr when calendar is '' (issue
-  #173).
+* Return the default values of dayofwk and dayofyr when calendar
+  is '' (issue #173).
 
 version 1.1.3 (release tag v1.1.3rel)
 =====================================
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,8 @@
+version <> (release tag v<>rel)
+=====================================
+* Return -1 from dayofwk and dayofyr when calendar is '' (issue
+  #173).
+
 version 1.1.3 (release tag v1.1.3rel)
 =====================================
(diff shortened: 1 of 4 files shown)


jswhit:
Looks good - ready to merge when failing tests are fixed.

PR #176, opened 2020-05-25
Add num2date_exact function to decode times exactly using timedelta addition

### LINKED ISSUE #171: Decoding times in num2date exactly with timedelta arithmetic
[Timedelta arithmetic in cftime uses integer arithmetic](https://github.com/Unidata/cftime/blob/b1a64bfc7fa24e7543cf73c5e30635b3eb0a9eb6/cftime/_cftime.pyx#L1745-L1758), which is precise down to the microsecond.  If passed an array of integers, or an array of floats that can safely be cast to integers, would it be possible for `num2date` to default to decoding times using that instead of the current imprecise floating point method?  This would be a helpful improvement to the user experience in downstream libraries like xarray, where microsecond noise introduced in the times by floating point arithmetic can interfere with automatic alignment.  See also other issues in this repo related to decoding precision, e.g. https://github.com/Unidata/cftime/issues/78, https://github.com/Unidata/cftime/issues/134.

[This gist](https://gist.github.com/spencerkclark/e8f9aed41acd8f524cef38706b3622b5) is a simplified sketch of how this code path could be implemented (I'm happy to develop it further into a PR if it seems promising).  The idea would be to try to use this method first, and fall back to using the existing floating point method if it wasn't possible.  The only reservation I would have about this is that it is about 9x slower than decoding times using floats.  In my use-cases, however, this slowdown would gladly be accepted if it meant times would be decoded exactly, but perhaps others decoding more than 10000 times at a time would find this frustrating.  

Are there any other reasons I'm not thinking of for why this might not be a good idea?

spencerkclark:
Closes #171.

This is a cleaned up implementation of `num2date_exact` illustrated in #171, complete with tests and support for masked input arrays.

I did a little more profiling.  It happens that the speed depends on how distant the times are from the reference date.  My initial tests were fairly extreme -- I was testing with timedeltas on the order of a million days, which resulted in some dates in year 4000 or higher.  For more typical use-cases, this new method is actually faster than the old method (I'm not sure if this changes the perspective on the default):

```
In [1]: import cftime

In [2]: import numpy as np

In [3]: UNITS = "microseconds since 1900-01-01"

In [4]: CALENDAR = "proleptic_gregorian"

In [5]: times = np.random.randint(0, 86400000000 * 100000, size=(10000, ))

In [6]: %timeit cftime.num2date(times, UNITS, CALENDAR)
97.6 ms ± 1.29 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [7]: %timeit cftime.num2date_exact(times, UNITS, CALENDAR)
83.2 ms ± 882 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [8]: cftime.num2date_exact(times, UNITS, CALENDAR)
Out[8]:
array([cftime.DatetimeProlepticGregorian(2082-11-09 05:54:21.915404),
       cftime.DatetimeProlepticGregorian(1939-01-12 18:24:47.502144),
       cftime.DatetimeProlepticGregorian(2072-01-05 15:25:01.058681), ...,
       cftime.DatetimeProlepticGregorian(2079-10-28 20:44:07.104899),
       cftime.DatetimeProlepticGregorian(1932-11-23 18:29:45.244291),
       cftime.DatetimeProlepticGregorian(1916-03-13 14:29:15.970997)],
      dtype=object)
```

This is obviously a fairly important part of cftime.  I think I got to everything in the test coverage, but please let me know if there's anything missing.  Happy to iterate as long as needed; I'm in no rush to get this in.

Note this leverages a version of the [`xarray.coding.times.cast_to_int_if_safe`](https://github.com/pydata/xarray/blob/d1f7cb8fd95d588d3f7a7e90916c25747b90ad5a/xarray/coding/times.py#L336-L340) function, which I think @shoyer added to xarray a while ago.  Is it ok if we use that here?


jswhit:
This looks fantastic @spencerkclark, thank you.
I suggest we get this in a 1.1.4 release, and make it the default if all goes well for the release after.

jswhit:
Perhaps a better name would be `num2date_int` (then the existing implementation can be changed to `num2date_float`, and `num2date` can point to `num2date_float` for now, and switch to `num2date_int` in the next release).

cftime/_cftime.pyx
```diff
@@ -396,6 +397,190 @@ OverflowError in python datetime, probably because year < datetime.MINYEAR"""
...
+
+    scaled_times = cast_to_int_if_safe(scaled_times)
+    if not np.issubdtype(scaled_times.dtype, np.integer):
+        warnings.warn(
+            "times must have integer dtype or be able to be safely cast to an "
+            "integer dtype to be decoded exactly.  Falling back to the "
+            "inexact version of the num2date function.")
+        return num2date(times, units, calendar, only_use_cftime_datetimes, only_use_python_datetimes)
+
+    shape = scaled_times.shape
```
jswhit, line 580: `shape` is never used?

spencerkclark: Indeed, thanks!

spencerkclark:
> Perhaps a better name would be `num2date_int` (then the existing implementation can be changed to `num2date_float`, and `num2date` can point to `num2date_float` for now, and switch to `num2date_int` in the next release).

Thanks @jswhit I think that is a good idea -- I updated things accordingly.  

If @shoyer is ok with us using [`xarray.coding.times.cast_to_int_if_safe`](https://github.com/pydata/xarray/blob/d1f7cb8fd95d588d3f7a7e90916c25747b90ad5a/xarray/coding/times.py#L336-L340) then I think this may be ready to go.  I guess beyond just that function, the approach I've taken here is largely inspired by [how times are decoded in xarray with pandas](https://github.com/pydata/xarray/blob/a2e9804d372a2e8a0d83904544b1d8421bfa2b44/xarray/coding/times.py#L125-L132), which seems to work pretty well.

jswhit:
I updated the `date2num` function to to use the UNIT_CONVERSION_FACTORS

jswhit:
> Perhaps a better name would be `num2date_int` (then the existing implementation can be changed to `num2date_float`, and `num2date` can point to `num2date_float` for now, and switch to `num2date_int` in the next release).

Changed my mind on this - let's make it the default now as long as all tests pass.  That way people will actually use it and we'll get more feedback.

cftime/_cftime.pyx
```diff
@@ -412,6 +461,189 @@ OverflowError in python datetime, probably because year < datetime.MINYEAR"""
...
+
+    scaled_times = cast_to_int_if_safe(scaled_times)
+    if not np.issubdtype(scaled_times.dtype, np.integer):
+        warnings.warn(
+            "times must have integer dtype or be able to be safely cast to an "
+            "integer dtype to be decoded exactly.  Falling back to the "
+            "inexact version of the num2date function.")
+        return num2date(times, units, calendar, only_use_cftime_datetimes, only_use_python_datetimes)
+
+    deltas = scaled_times.astype("timedelta64[us]").astype(timedelta)
```
jswhit, line 644: What does this line do? (please add comment)

cftime/_cftime.pyx
```diff
@@ -412,6 +461,189 @@ OverflowError in python datetime, probably because year < datetime.MINYEAR"""
...
+    factor = UNIT_CONVERSION_FACTORS[unit]
+    scaled_times = factor * times
+
+    scaled_times = cast_to_int_if_safe(scaled_times)
+    if not np.issubdtype(scaled_times.dtype, np.integer):
+        warnings.warn(
+            "times must have integer dtype or be able to be safely cast to an "
+            "integer dtype to be decoded exactly.  Falling back to the "
+            "inexact version of the num2date function.")
+        return num2date(times, units, calendar, only_use_cftime_datetimes, only_use_python_datetimes)
```
jswhit, line 642: Needs to be `num2date_float`

cftime/_cftime.pyx
jswhit, line 644: What does this line do? (please add comment)

cftime/_cftime.pyx
```diff
@@ -304,6 +297,62 @@ def num2date(times,units,calendar='standard',\
...
+    objects if `calendar='proleptic_gregorian'`, or
+    `calendar='standard'` or `'gregorian'`
+    and the date is after the breakpoint between the Julian and
+    Gregorian calendars (1582-10-15). Otherwise, they are ctime.datetime
+    objects which support some but not all the methods of native python
+    datetime objects. The datetime instances
+    do not contain a time-zone offset, even if the specified `units`
+    contains one.
+    """
+    return num2date_float(
```
jswhit, line 312: Let's go ahead, bite the bullet,  and use `numdate_int` as long as all the tests pass.

jswhit:
@spencerkclark I'd like to get this merged, can you take a look at my review?

spencerkclark:
Thanks for the careful review and for the ping @jswhit; sorry for letting this slip.  I'll try and fix this up by the end of the week.

cftime/_cftime.pyx
```diff
@@ -412,6 +461,189 @@ OverflowError in python datetime, probably because year < datetime.MINYEAR"""
...
+        basedate = to_calendar_specific_datetime(basedate, calendar)
+
+    if unit not in UNIT_CONVERSION_FACTORS:
+        raise ValueError("Unsupported time units provided, {!r}.".format(unit))
+
+    if unit in ["months", "month"] and calendar != "360_day":
+        raise ValueError("Units of months only valid for 360_day calendar.")
+
+    factor = UNIT_CONVERSION_FACTORS[unit]
+    scaled_times = factor * times
```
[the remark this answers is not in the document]
spencerkclark: Good idea to cast float `times` to `np.longdouble` prior to scaling.  I also added some code to `cast_to_int_if_safe` which prevents integer casting if the values are outside the longdouble integer range on the particular platform.  Let me know if that looks correct.

I make sure to cast any integer inputs to `int64` before doing any multiplication too.

> without casting to an array this will fail with list input.

Thanks for pointing this out -- I added a test case for list input and cast `times` as an array before doing anything else.

cftime/_cftime.pyx
spencerkclark: Indeed, thanks.

cftime/_cftime.pyx
spencerkclark: Sure thing; it looks like they all do now.

spencerkclark:
Thanks again @jswhit -- when you get a chance, I think things should be ready for another pass.

jswhit:
Looks good @spencerkclark - merging now

Issue #109, opened 2019-01-18
Question regarding precision of datetime arithmetic


spencerkclark:
We are working on a cftime-compatible version of pandas' resample functionality in xarray (pydata/xarray#2593).  In order to accomplish this, we are making heavy use of the datetime arithmetic functionality in cftime.  There are two contexts where this occurs:
- Adding a `datetime.timedelta` object do a `cftime.datetime` object to produce another `cftime.datetime` object
- Taking the difference between two `cftime.datetime` objects to produce a `datetime.timedelta` object.

It is my understanding that in the first context, [integer arithmetic is used](https://github.com/Unidata/cftime/blob/ca6ddce6fbfddfb769de6e79947a4af4b471fc93/cftime/_cftime.pyx#L1644-L1657), so the result is microsecond-precise; in the second context, however, things follow a different code path ([using `date2num`](https://github.com/Unidata/cftime/blob/ca6ddce6fbfddfb769de6e79947a4af4b471fc93/cftime/_cftime.pyx#L1350)) and arithmetic is not necessarily exact (in my tests on a Mac this issue seems to have mostly gone away in cftime version 1.0.3.4; however @jwenfai has found on Windows that issues persist).

We are finding that having an exact result for the difference between two datetimes is important for our new use-case.  A potential solution one might think of trying is to try to write a function that computes exact datetime differences.

Here is a potential function I came up with.  Do you think that this would be a safe workaround for this issue, or might there be issues I'm not anticipating?  Thanks for your help.
```python
from datetime import timedelta

def exact_cftime_datetime_difference(a, b):
    """Exact computation of b - a

    Assumes:

        a = a_0 + a_m
        b = b_0 + b_m

    Here a_0, and b_0 represent the input dates rounded 
    down to the nearest second, and a_m, and b_m represent
    the remaining microseconds associated with date a and
    date b.

    We can then express the value of b - a as:

        b - a = (b_0 + b_m) - (a_0 + a_m) = b_0 - a_0 + b_m - a_m
    
    By construction, we know that b_0 - a_0 must be a round number
    of seconds.  Therefore we can take the result of b_0 - a_0 using
    ordinary cftime.datetime arithmetic and round to the nearest
    second.  b_m - a_m is the remainder, in microseconds, and we
    can simply add this to the rounded timedelta.

    Parameters
    ----------
    a : cftime.datetime
        Input datetime
    b : cftime.datetime
        Input datetime

    Returns
    -------
    datetime.timedelta
    """
[...]

PR #178, opened 2020-06-21
Make cftime.datetime - cftime.datetime -> timedelta microsecond-exact

spencerkclark:
Currently subtracting a date from another to produce a timedelta is not microsecond-exact:

```
In [1]: import cftime

In [2]: cftime.DatetimeNoLeap(2000, 1, 2, 0, 0, 0, 5) - cftime.DatetimeNoLeap(2000, 1, 2)
Out[2]: datetime.timedelta(microseconds=8)
```

I think this is actually pretty straightforward to fix using some code that already exists in cftime.  I'm drawing inspiration here from [the way this kind of operation is done in Python's datetime object](https://github.com/python/cpython/blob/a4c09560eaa1e3b9733c7e63d570b6fa7626724c/Lib/datetime.py#L2084-L2097).  As I understand it, we can simply use `_IntJulianDayFromDate` as something essentially equivalent to `datetime.datetime.toordinal`, and everything else should basically be the same.

The test confirms the example above is fixed for all calendars, and all other existing tests of timedelta arithmetic pass.  @jswhit I think this should enable the exact version of `date2num` you described in https://github.com/Unidata/cftime/issues/171#issuecomment-636201592.  I could try to add that here or in another PR.  As you note, it would be awesome to be able to round-trip dates without loss of precision.

xref: #109 (this would make the function described there no longer necessary)

### DIFF
--- a/cftime/_cftime.pyx
+++ b/cftime/_cftime.pyx
@@ -1280,7 +1280,6 @@ _converters = {}
 for calendar in _calendars:
     _converters[calendar] = utime("seconds since 1-1-1", calendar)
 
-
 @cython.embedsignature(True)
 cdef class datetime(object):
     """
--- a/cftime/_cftime.pyx
+++ b/cftime/_cftime.pyx
@@ -1280,6 +1280,7 @@ _converters = {}
 for calendar in _calendars:
     _converters[calendar] = utime("seconds since 1-1-1", calendar)
 
+
 @cython.embedsignature(True)
 cdef class datetime(object):
     """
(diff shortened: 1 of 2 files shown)


jswhit:
Wow - very simple and elegant solution!   I can merge once you resolve the conflicts.

spencerkclark:
Thanks @jswhit -- indeed I was pleasantly surprised to find that I could reuse the `_IntJulianDayFromDate` function, which was already so nicely implemented for the original version of the calculation!  

Merge conflicts have been resolved; I think things should be ready to go.

jswhit:
Merging now.  I will wait a week or so and then push another release.

jswhit:
Also, it would be good to run the xarray test suite to make sure this has not broken anything before a new release

spencerkclark:
> Oops - we forgot the Changelog entry.

Ah yes, thanks for reminding me -- see #179.

> Also, it would be good to run the xarray test suite to make sure this has not broken anything before a new release

Yes, I've confirmed that the xarray test suite passes with the latest updates.  I'm looking forward to eventually being able to remove some of the tolerance logic we have in our cftime round-tripping tests (we'll still support older versions of cftime for some time).

PR #180, opened 2020-06-24
have utime.num2date/utime.date2num call module level functions

jswhit:
JulianDayFromDate, DateFromJulianDay no longer used internally.
Bump version number to 1.2.0.


jswhit:
@spencerkclark - your latest improvements to date2num and num2date work so well that I decided to go ahead and use them everywhere.  Would you mind running this with the xarray test suite?  If it all looks good, I will merge and create a 1.2.0 release.

cftime/_cftime.pyx
```diff
@@ -121,344 +121,163 @@ def _datesplit(timestr):
...
-            n += 1
-        if ismasked: # convert to masked array if input was masked array
-            dates = np.array(dates)
-            dates = np.ma.masked_where(dates==None,dates)
-            if isscalar:
-                return dates[0]
-            else:
-                return np.reshape(dates, shape)
+                #times.append( (td/timedelta(microseconds=1)) / factor )
+                # this works for python 2.7
```
spencerkclark, line 257: Could we switch to the Python 3 specific logic since we are now dropping Python 2.7 support?

cftime/_cftime.pyx
```diff
@@ -121,344 +121,163 @@ def _datesplit(timestr):
...
+    if np.ma.isMA(dates) and np.ma.is_masked(dates):
+        mask = dates.mask
+        ismasked = True
+    if isscalar:
+        dates = np.array([dates])
+    else:
+        dates = np.array(dates)
+        shape = dates.shape
+    times = []
+    for date in dates.flat:
```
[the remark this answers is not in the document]
spencerkclark: Somewhat related to this, I think there may still be something amiss with respect to masked arrays:
```
In [1]: import numpy as np

In [2]: import cftime

In [3]: times = np.array([cftime.DatetimeNoLeap(2000, 1, 1), cftime.DatetimeNoLeap(2000, 1, 2)])

In [4]: times = np.ma.masked_array(times, mask=[False, True])

In [5]: times
Out[5]:
masked_array(data=[cftime.DatetimeNoLeap(2000, 1, 1, 0, 0, 0), --],
             mask=[False,  True],
       fill_value='?',
            dtype=object)

In [6]: cftime.date2num(times, units="days since 2000-01-01")
Out[6]:
masked_array(data=[0., 1.],
             mask=False,
       fill_value=1e+20)
```
I would expect the last value of the result to remain masked.

test/test_cftime.py
```diff
@@ -777,10 +774,37 @@ def test_tz_naive(self):
...
         dates = num2date(times, units=units, calendar='standard',\
         only_use_cftime_datetimes=False, only_use_python_datetimes=True)
-        assert((dates==[datetime(1848, 1, 17, 6, 0, 0, 40),None]).all())
+        test = dates == np.ma.masked_array([datetime(1848, 1, 17, 6, 0, 0, 40), None],mask=[0,1])
+        assert(test.all())
         dates = num2date(times, units=units, calendar='standard')
-        assert((dates==[cftime.DatetimeGregorian(1848, 1, 17, 6, 0, 0),None]).all())
-        
+        assert(str(dates)=='[cftime.DatetimeGregorian(1848, 1, 17, 6, 0, 40) --]')
+#  check that time range of 200,000 + years can be represented accurately
```
spencerkclark, line 781: Thanks for adding this test!

test/test_cftime.py
```diff
@@ -1786,32 +1810,32 @@ def test_num2date_int_only_use_python_datetimes_invalid_basedate(
...
-    num2date_int(numeric_times, units=units, calendar=artificial_calendar)
+    num2date(numeric_times, units=units, calendar=artificial_calendar)
 
 
-def test_num2date_int_uncastable_values(calendar):
-    units = "days since 2000-01-01"
-    numeric_times = np.array([1.0, np.pi])
-    with pytest.warns(UserWarning, match="Falling back to the older inexact"):
-        num2date_int(numeric_times, units=units, calendar=calendar)
+#def test_num2date_uncastable_values(calendar):
```
spencerkclark, line 1831: It's probably ok to delete this test now.

test/test_cftime.py
```diff
@@ -1821,30 +1845,31 @@ def test_num2date_int_masked_array(calendar):
...
 
 
-def test_num2date_int_out_of_range():
+def test_num2date_out_of_range():
     numeric_times = 12 * np.array([200000, 400000, 600000])
     units = "months since 2000-01-01"
-    with pytest.warns(UserWarning, match="Falling back to the older inexact"):
-        num2date_int(numeric_times, units=units, calendar="360_day")
+    with pytest.raises(OverflowError, match="time values outside range of 64 bit signed integers"):
+    #with pytest.warns(UserWarning, match="Falling back to the older inexact"):
```
spencerkclark, line 1856: Should also be safe to remove now.

cftime/_cftime.pyx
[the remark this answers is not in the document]
jswhit: Thanks for catching that __repr__ bug!  Fixed now, along with the masked array bug.  
Getting rid of the loop in date2num is going to be trickier, since there are so may checks/if tests in there.  Going to leave that for another PR.

test/test_cftime.py
jswhit: Done

cftime/_cftime.pyx
jswhit: Done

Issue #198, opened 2020-08-25
cftime.datetime does not allow addition


mcgibbon:
`cftime.datetime` returns an object of type `cftime._cftime.datetime`, which cannot be added to or subtracted from using a `timedelta`. Based on its `calendar` keyword argument, I would expect it to return a subclass such as `DatetimeJulian` which has arithmetic defined.

```
Python 3.7.4 (default, Aug 13 2019, 15:17:50)
[Clang 4.0.1 (tags/RELEASE_401/final)] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import cftime
>>> dt = cftime.datetime(2020, 1, 1, calendar="julian")
>>> from datetime import timedelta
>>> dt += timedelta(hours=1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for +=: 'cftime._cftime.datetime' and 'datetime.timedelta'
>>> cftime.__version__
'1.2.1'
```

spencerkclark:
> Based on its `calendar` keyword argument, I would expect it to return a subclass such as `DatetimeJulian` which has arithmetic defined.

I would be +1 on this.  It's often quite useful to be able to construct datetimes starting from a string calendar name rather than an explicit `cftime` type, which extending the constructor of `cftime.datetime` in this way would enable.

jswhit:
Pull request would be welcomed.

jswhit:
PR #199 implements this by changing the name of the base class `datetime` to `_datetime` and making `datetime` a factory function that returns calendar-specific datetime instances.

jswhit:
PR #201 is an alternate fix that does not require changing the name of the base class.  It does require changing the default calendar from `standard` to `""`.  Since much of the code is moved from cython to pure python for this solution, it is quite a bit slower.  More optimization is needed before it can be merged.

jswhit:
PR #202 is yet another alternate fix that does not require changing the name of the base class, nor does it degrade performance by migrating the base class to python.  It creates 'calendar-aware' instances of the base-class directly using the calendar kwarg to __init__ and a bunch if if/else statements in `__add__` and `__sub__`.  The sub-classes are still there, although they no longer over-ride any methods of the base class.  This should work fine as long as no one is checking for sub-classes.

jswhit:
[...]

PR #199, opened 2020-08-27
fix for issue #198

jswhit:


spencerkclark:
Thanks for working on this @jswhit; let me think about it a little more.  If possible I would like to retain the base class as public API, since it is useful for instance checks in functions that accept cftime dates of any calendar type.  One solution might be to do something similar to how pandas handles creating indexes of different types from the generic `pandas.Index` constructor, e.g. by defining a `__new__` method, but I'm not sure if that would work in Cython.

mcgibbon:
I was thinking about this and I'd second @spencerkclark's suggestion that this be done using a `__new__` method or similar approach that lets you have `cftime.datetime` still be a class, but also initialize calendar-specific classes. Great that there's already an example in Pandas. If Cython doesn't allow it, you could always write the `cftime.datetime` class itself as a pure Python class.

mcgibbon:
This could be possible by making the `datetime` function a "fake" class which pretends it's the datetime base class for the purposes of `isinstance` and `issubclass`, by setting a `__class__` attribute on the function (or one of the other approaches suggested [here](https://stackoverflow.com/questions/6803597/how-to-fake-type-with-python)).

jswhit:
If you set `calendar=''` you get an instance of the base class.

mcgibbon:
Yes, but ideally user code which currently says `isinstance(my_datetime, cftime.datetime)` shouldn't be broken by this change.

jswhit:
Right, I see your point

jswhit:
the base class is now part of the public API, but it's called `datetime_base`.  This means user code will have to be changed to use `isinstance(my_datetime, cftime.datetime_base)`.  Not ideal.

mcgibbon:
I believe you can create a class which has a __call__ method which is what your PR has as the `datetime` function, and a __class__ property attribute which is `datetime_base`. It would take some checking/writing tests that the behavior is as expected for isinstance and issubclass.

jswhit:
Yes, but then users will still have to  use `isinstance(my_datetime, cftime.datetime_base)`.

jswhit:
Before I merge this, I would like to know from the xarray team whether changing the name of the base class from `cftime.datetime` to `cftime.datetime_base` will be a problem.  @spencerkclark ?

spencerkclark:
Thanks @jswhit -- I am indeed still hesitant to merge this.  This would be a breaking change for xarray, as we use the base class in several places and would need to add version-dependent import logic to retain backwards compatibility.

mcgibbon:
> Yes, but then users will still have to use isinstance(my_datetime, cftime.datetime_base).

The idea is that you can do some fancy Python magic so that `isinstance(my_datetime, cftime.datetime)` works even though `cftime.datetime` is either technically a function, or a class whose `__new__` method may actually return a subclass.

I could try to work out a solution that does this.

jswhit:
That would be the ideal solution, but I have not been able to make this work in cython.

mcgibbon:
Yes, it appears that Cython doesn't allow any of the ways that Python would allow doing this, and Cython classes also can't inherit from Python classes.

The only "solution" I can think of is if `Calendar` were its own class distinct from `datetime`, with its own subclasses, so that `cftime` can provide a single `datetime` class used for all calendars. This would be a significant change to how the package works, but it could possibly maintain backwards compatibility if you retain dummy datetime subclasses that inherit from the datetime base and only modify the init routine to set their specific calendar. This kind of sidesteps the calendar-class-subclass-from-datetime-while-also-datetime-can-make-calendar-specific-classes issue because you no longer have calendars as "datetime"-like classes.

Unless you want to make a larger change like that, I'd say the current "bug" is better than breaking `isinstance(my_datetime, cftime.datetime)`.

jswhit:
If you can figure out how to do it in pure python, we can make the datetime base class a python class instead of a cython extension object.  I don't think there is a much of a performance hit for this.

davidhassell:
cf-python also explicitly uses the current `cftime.datetime`, but I'm OK with the proposed new API. I do not (at the moment)  provide backwards compatible support for dependency libraries, so this is no problem for me to change the cf-python code to match. Thanks.

jswhit:
I have created a `pure_python` branch which doesn't use cython. It's slow, but can be used to prototype solutions to the problem raised here.  If we have a working solution, the performance sensitive parts can be migrated back to cython.

jswhit:
@spencerkclark - instead of version checking to maintain backward compatability, could you do something like this?

```python
from cftime import datetime
try:
    from cftime import datetime_base
except ImportError:
    datetime_base = datetime
d = datetime(1962,10,27,12,calendar='')
print(type(d))
print(isinstance(d,datetime_base))
```

spencerkclark:
Sorry @jswhit, yes, we can work around things that way too.  That's similar to how we handled things when `cftime` was split out of `netCDF4`.  Give me a chance to make a PR to xarray to preemptively fix this, as if you merge this right away our upstream-dev build on CI, which tests against the master version of cftime among other packages, will break.  I'll ping you when things are ready.

spencerkclark:
Ok, I forgot, this also will lead to issues in `nc_time_axis`, which may take longer to sort out.  Not because those issues are hard to fix technically -- the same solution will work -- but we'll need to get a PR merged there too.  See https://github.com/SciTools/nc-time-axis/pull/49.

mcgibbon:
> I'd say the current "bug" is better than breaking `isinstance(my_datetime, cftime.datetime)`.

Again, I don't think fixing this issue is worth breaking user code. A full fix (full in the sense that it allows the same `datetime` class to give both `isinstance` and initialization functionality, as the built-in class does) could mean creating a unified `cftime.datetime` class which uses a `Calendar` object internally to implement the specific choice of calendar. If you later implemented that full fix, it would require breaking user code again.

It's up to you package developers as to what you want to do, I just wanted to be clear as the one who brought up the issue.

spencerkclark:
While I have done the work to sketch out what it would take to address this in the primary downstream packages, after thinking about this some more, I tend to agree with @mcgibbon.  I don't know the extent to which others are relying on the name of this base class.  I would probably prefer to leave this be for the time being.

Sorry for the noise @jswhit; I appreciate the effort you made to try and address this particular issue.  It's too bad there is not an easy way to do this in Cython.  @mcgibbon's idea for unifying the datetime classes is interesting, but indeed it would be a big undertaking, so it would take someone with a fair amount of time on their hands to address (in other words, please don't feel compelled to do it!).

jswhit:
I still think that the benefits of this change outweigh the potential compatibilty issues.  We could make this a 2.0 release and clearly note that there are backwards incompatible API changes.  I'll defer to @spencerkclark and @mcgibbon on this though - since they are the ones most affected.

jswhit:
The `pure_python` branch contains a solution that does not require renaming the base class.  It does require changing the default calendar in `cftime.datetime` from `standard` to ````, however.    The solution implemented uses __new__ to return the calendar specific instance - since you can't do this in cython I had to move the base class to python.  This has a cascading effect, since cython extension classes cannot inherit from python classes.  The result is that it works, but it's much slower (`num2date` is about 10 times slower).  I'm sure there are ways to optimize things further, so it remains to be seen how much of that performance can be recovered.

PR #201, opened 2020-09-25
alternate fix for issue #198

jswhit:
that does not require changing the name of the base class (`cftime.datetime`).
Degrades performance however, since most of the code had to be moved into pure python.

@spencerkclark - when you get a chance, please see if all the xarray tests pass with the `pure_python` branch.  I'd also be curious to know if you notice the slowness.


mcgibbon:
Thank you for the work on this!

PR #205, opened 2020-11-05
switch default calendar in date2num to None

jswhit:
which causes date2num to use the calendar associated with the first input datetime instance (proleptic_gregorian if datetime instance is native python).  `calendar=""` has the same effect as `calendar=None`.


jswhit:
closes issue #206

PR #217, opened 2020-12-27
print more informative error msg when importing from install dir

jswhit:


jswhit:
also fix coverage/coveralls step in workflow.  Update to cython language_level=3 in setup.py

PR #225, opened 2021-01-23
Further improvements to datetime roundtripping

spencerkclark:
I forget if there was a reason we did not do this before, but this applies an approach I took in xarray (https://github.com/pydata/xarray/pull/4684) to improve the accuracy of roundtripping `np.datetime64[ns]` dates to `cftime.datetime` objects. The changes here would mean that we'd do our absolute best to retain an integer dtype when encoding times through `date2num`.  

I've included a pretty challenging test that passes on my machine; ~~however it may not be a valid test on 32-bit systems.~~  Edit: I updated the PR for what I think should allow things to pass on Windows.

### DIFF
--- a/.github/workflows/miniconda.yml
+++ b/.github/workflows/miniconda.yml
@@ -40,7 +40,9 @@ jobs:
       run: |
         conda create --name TEST python=${{ matrix.python-version }} --file requirements.txt --file requirements-dev.txt
         source activate TEST
-        CYTHON_COVERAGE=1 pip install -v -e  . --no-deps --force-reinstall
+        # enabling coverage slows down the tests dramaticaly
+        #CYTHON_COVERAGE=1 pip install -v -e  . --no-deps --force-reinstall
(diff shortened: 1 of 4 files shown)


spencerkclark:
I'm not sure why the tests take so long in CI (previous runs took 3+ hours!).  The full test suite runs in about 25 seconds on my local machine.

jswhit:
I have no idea why the tests are so slow all of a sudden - they run super fast on my machine too.

jswhit:
I think it's the code coverage that's slowing down the CI tests.  I just disabled it in the CI workflow in your branch.

jswhit:
This is really nice @spencerkclark, now we are very close to microsecond roundtrip accuracy.  I think all this needs is a Changelog entry.

spencerkclark:
Ah, disabling the code coverage seemed to do it, thanks!  I went ahead and added a Changelog entry.  Let me know if you need anything else.

PR #227, opened 2021-01-30
prepare for v1.4.0 release

jswhit:
`cftime.datetime` now no longer uses calendar-specific sub-classes.  Legacy functions (including calendar-specific subclasses) now kept in a separate file (`_cftime_legacy.pyx`).


jswhit:
@spencerkclark  - would you mind running the xarray tests again on this branch before I make a release? I decided to remove the calendar-specific sub-classes from cftime.datetime and I want to make sure that doesn't break anything.  (`__add__` and `__sub__` used to return the calendar specific sub-classes)

spencerkclark:
That's nice, and it looks like you kept the ability to import the calendar-specific sub-classes, which we do still use in the xarray tests (we can probably migrate away from that at some point).  I ran the tests with this branch and everything passed.

jswhit:
thanks @spencerkclark. merging now.

spencerkclark:
> That's nice, and it looks like you kept the ability to import the calendar-specific sub-classes, which we do still use in the xarray tests (we can probably migrate away from that at some point). I ran the tests with this branch and everything passed.

@jswhit unfortunately I think the fact that the xarray test suite uses the legacy types prevented it from catching some (fairly serious) errors that arose when using the new universal `cftime.datetime` objects.  Perhaps making that change was a bit premature after all.  Would you be open to reverting that change and issuing a patch release?  Since this was my mistake, I'd be happy to take on the work of writing the PR to revert those additional changes.

spencerkclark:
xref: https://github.com/pydata/xarray/issues/4853

spencerkclark:
@jswhit many thanks for issuing the 1.4.1 release!  I missed it a few days ago because it did not show up in the "Releases" section on the GitHub repo homepage.  

Screen Shot 2021-02-05 at 2 53 36 PM

Is there a way to update that too?

PR #232, opened 2021-03-05
cftime.datetime support for __format__

jamesp:
This PR adds the `__format__` method to cftime datetime objects.  This allows the use of cftime.datetimes inline within format strings.  Such as:

```
dt = cftime.datetime(2020,1,1)
mystring = "the year is {0:%Y}".format(dt) # >>> the year is 2000
```

and with format strings:
```
mystring = f"the year is {dt:%Y}"  # >>> the year is 2000
```

Things brings formatting parity with `datetime.datetime`.  At present, code like the below will fail if cftime objects are returned, or succeed if datetime objects are returned:
```
dt = num2date(**kwargs)
# dt could be a datetime.datetime or cftime.datetime
print(f"the date is {dt:%Y-%m-%d}")  # currently fails if isinstance(a, cftime.datetime)
```

### DIFF
--- a/test/test_cftime.py
+++ b/test/test_cftime.py
@@ -1,5 +1,4 @@
 from __future__ import print_function
-from calendar import c
 
 import copy
 import operator
--- a/test/test_cftime.py
+++ b/test/test_cftime.py
@@ -1,4 +1,5 @@
 from __future__ import print_function
+from calendar import c
 
 import copy
 import operator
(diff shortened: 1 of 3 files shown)


jswhit:
Very nice  -thanks @jamesp!  Could you just add a Changelog entry please?

test/test_cftime.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import print_function
+from calendar import c
```
spencerkclark, line 2: Is this needed?

spencerkclark (commented):
Indeed, this is cool!

test/test_cftime.py
jamesp: No, that looks like a "helpful" addition from vscode auto-import... good spot thanks! will remove

jamesp:
Thanks @jswhit, @spencerkclark.  Errant import removed and changelog updated 👍

PR #238, opened 2021-03-30
change to proleptic_gregorian defaults (has_year_zero=True) 

jswhit:
to conform with upcoming CF v1.9 (https://github.com/cf-convention/cf-conventions/pull/315)

Also enables warning if calendar requested does not conform with CF v1.9


jswhit:
The CF calendar conventions were updated for version 1.9 (see https://github.com/cf-convention/cf-conventions/pull/331/files)

jswhit:
the relevant updates to CF are copied below:

```
The values currently defined for **`calendar`** are listed below. In all calendars except **`360_day`** and **`none`**, the lengths of the months are the same as in the Gregorian calendar for leap years and non-leap years. In the **`julian`** and the default **`standard`** mixed Gregorian/Julian calendar, dates in years before year 0 (i.e. before 0-1-1 0:0:0) are not allowed, and the year in the reference date/time of the units must not be negative. In these calendars, year zero has a special use to indicate a climatology (see <<climatological-statistics>>), but this use of year zero is deprecated. In other calendars, years before year 1 are allowed.


**`proleptic_gregorian`**:: A calendar with the Gregorian rules for leap-years extended to dates before 1582-10-15. All dates consistent with these rules are allowed, both before and after 1582-10-15 0:0:0.
```


To summarize, dates before year 0 are no longer allowed for `julian` or `standard` calendars, and the use of year zero is deprecated.  year zero and negative years are allowed for `proleptic_gregorian`.  In addition the `gregorian` calendar name is deprecated (currently a synonym for `standard`.

jswhit:
CF version 1 is [released](http://cfconventions.org/cf-conventions/cf-conventions#calendar)

Issue #243, opened 2021-05-25
Add support to release aarch64 wheels


odidev:
Problem
---------
On aarch64, ‘pip install cftime’ builds the wheels from source code and then installs it. It requires the user to have a development environment installed on his system. Also, it takes some more time to build the wheels than downloading and extracting the wheels from pypi.

Resolution
-----------
On aarch64, ‘pip install cftime’ should download the wheels from pypi

@jswhit Please let me know your interest in releasing aarch64 wheels. I can help with this.

jswhit:
The wheel building repo is at http://github.com/MacPython/cftime-wheels.  It uses github actions (yaml in .github/workflows/build-windows-wheels.yml) to build ManyLinux wheels using multibuild.  Feel free to create a PR there to add support for building aarch64 wheels.

jswhit:
It also builds macos x wheels - is this the architecture for the new M1 macs?

odidev:
I have raised the PR to build the wheel for linux aarch64 using multibuild.
**PR Link** - https://github.com/MacPython/cftime-wheels/pull/6

jswhit:
MacPython/cftime-wheels#6 was merged and aarch64 wheels are now up on pypi.

PR #244, opened 2021-05-25
Add linux aarch64 wheel build support

odidev:
Add linux aarch64 wheel build support.
Related to https://github.com/Unidata/cftime/issues/243 @jswhit Could you please review this PR?

### DIFF
--- a/.github/workflows/miniconda.yml
+++ b/.github/workflows/miniconda.yml
@@ -83,3 +83,43 @@ jobs:
         #rm -f docs/build/html/_static/underscore-*.js
         #rm -f docs/build/html/.buildinfo
         #sh ./ci/deploy_docs.sh
+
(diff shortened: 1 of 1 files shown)


jswhit:
Moved to https://github.com/MacPython/cftime-wheels/pull/5

odidev:
> We just need to run the tests here on aarch64, the wheels should be built in the cftime-wheels repo. Can aarch64 just be added to the platform matrix and run just if os === ubuntu? (EDIT: this seems to work, see #245)

I have checked the [work](https://github.com/Unidata/cftime/pull/245). It is running for x86_64 and not for AArch64.GHA provides only x86_64 builders. We need to do something like: https://github.com/marketplace/actions/run-on-architecture#supported-platforms.

PR #246, opened 2021-05-28
Add support for 'common_year' and 'common_years' units

andersy005:
This PR is my attempt at fixing #5. As discussed in #5, the only supported calendars for year units are `365_day` and `noleap`. 

```python 
In [1]: import cftime

In [2]: values = [1, 2, 3]

In [3]: units = 'common_year since 0000-01-01 0:0:0'

In [4]: cftime.num2date(values, units, 'noleap')
Out[4]: 
array([cftime.DatetimeNoLeap(1, 1, 1, 0, 0, 0, 0, has_year_zero=True),
       cftime.DatetimeNoLeap(2, 1, 1, 0, 0, 0, 0, has_year_zero=True),
       cftime.DatetimeNoLeap(3, 1, 1, 0, 0, 0, 0, has_year_zero=True)],
      dtype=object)

In [5]: cftime.num2date(values, units, '365_day')
Out[5]: 
array([cftime.DatetimeNoLeap(1, 1, 1, 0, 0, 0, 0, has_year_zero=True),
       cftime.DatetimeNoLeap(2, 1, 1, 0, 0, 0, 0, has_year_zero=True),
       cftime.DatetimeNoLeap(3, 1, 1, 0, 0, 0, 0, has_year_zero=True)],
      dtype=object)
```

Using `common_year` or `years` units with other calendars results in failures as intended:

```python
n [6]: cftime.num2date(values, units, 'standard')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-6-5819edfbac07> in <module>
----> 1 cftime.num2date(values, units, 'standard')

~/devel/unidata/cftime/src/cftime/_cftime.pyx in cftime._cftime.num2date()
    488     if has_year_zero is None:
    489         has_year_zero = _year_zero_defaults(calendar)
--> 490     basedate = _dateparse(units,calendar=calendar,has_year_zero=has_year_zero)
    491 
    492     can_use_python_datetime=_can_use_python_datetime(basedate,calendar)

~/devel/unidata/cftime/src/cftime/_cftime.pyx in cftime._cftime._dateparse()
     99             raise ValueError("'months since' units only allowed for '360_day' calendar")
    100         if units in year_units and calendar not in {'365_day', 'noleap'}:
--> 101             raise ValueError("'%s' units only allowed for '365_day' and 'noleap' calendars" % units)
    102         else:
    103             raise ValueError(

ValueError: 'common_year' units only allowed for '365_day' and 'noleap' calendars
```

@kmpaul, let me know if this PR fully addresses #5.

### DIFF
--- a/src/cftime/_cftime.pyx
+++ b/src/cftime/_cftime.pyx
@@ -22,6 +22,7 @@ min_units =      ['minute', 'minutes', 'min', 'mins']
 hr_units =       ['hour', 'hours', 'hr', 'hrs', 'h']
 day_units =      ['day', 'days', 'd']
 month_units =    ['month', 'months'] # only allowed for 360_day calendar
+year_units =     ['common_year', 'years'] # only allowed for 365_day and noleap calendars
 _units = microsec_units+millisec_units+sec_units+min_units+hr_units+day_units
 # supported calendars. Includes synonyms ('standard'=='gregorian',
 # '366_day'=='all_leap','365_day'=='noleap')
@@ -93,9 +94,11 @@ def _dateparse(timestr,calendar,has_year_zero=None):
     if has_year_zero is None:
         has_year_zero = _year_zero_defaults(calendar)
     (units, isostring) = _datesplit(timestr)
-    if not ((units in month_units and calendar=='360_day') or units in _units):
+    if not ((units in month_units and calendar=='360_day') or (units in year_units and calendar in {'365_day', 'noleap'}) or units in _units):
         if units in month_units and calendar != '360_day':
             raise ValueError("'months since' units only allowed for '360_day' calendar")
+        if units in year_units and calendar not in {'365_day', 'noleap'}:    
+            raise ValueError("'%s' units only allowed for '365_day' and 'noleap' calendars" % units) 
         else:
             raise ValueError(
             "units must be one of 'seconds', 'minutes', 'hours' or 'days' (or singular version of these), got '%s'" % units)
@@ -320,7 +323,10 @@ UNIT_CONVERSION_FACTORS = {
     "days": 86400 * 1000000,
     "d": 86400 * 1000000,
     "month": 30 * 86400 * 1000000,  # Only allowed for 360_day calendar
-    "months": 30 * 86400 * 1000000
+    "months": 30 * 86400 * 1000000,
+    "common_year": 365 * 86400 * 1000000, # Only allowed for 365_day and no_leap calendars
+    "years": 365 * 86400 * 1000000 # Only allowed for 365_day and no_leap calendars
+    
 }
 
 DATE_TYPES = {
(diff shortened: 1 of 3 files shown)


jswhit:
According to the [udunits docs](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch04s04.html) a `common_year` is 365 days, but a `year` is 365.242198781 days.  In addition, a `leap_year` is 366 days, a `julian_year` is 365.25 days and a `gregorian_year` is 365.2425 days.  

So, I think there may some ambiguity in allowing 'years since' - does it mean 365 or 365.242198781 days? I guess that's why they advise against using years as a unit.

jswhit:
seems like if it works for 'noleap',  'years since' should also be supported for 'all_leap' and '360_day' calendars

jswhit:
Allowing 'year' units is a rabbit hole I don't think we really want to go all the way down.  The only examples of netcdf files with that use the 'common_years' time units in the wild that I can find seem to be output from the ice-sheet model in CESM, which uses the '365_day' calendar.  I'm willing to merge this PR to deal with this particular use-case, but after thinking about this some more I don't think enabling this for other calendars is a good idea.

jswhit:
Needs a Changelog entry

andersy005:
> So, I think there may some ambiguity in allowing years since - does it mean 365 or 365.242198781 days? I guess that's why they advise against using years as a unit.
> Allowing 'year' units is a rabbit hole I don't think we really want to go all the way down.

Thank you for your feedback, @jswhit! I concur. I'm going to remove the `years` units and only keep `common_year` units and will update the changelog.

src/cftime/_cftime.pyx
```diff
@@ -22,6 +22,7 @@ min_units =      ['minute', 'minutes', 'min', 'mins']
 hr_units =       ['hour', 'hours', 'hr', 'hrs', 'h']
 day_units =      ['day', 'days', 'd']
 month_units =    ['month', 'months'] # only allowed for 360_day calendar
+year_units =     ['common_year', 'years'] # only allowed for 365_day and noleap calendars
```
jswhit, line 25: maybe replace 'years' with 'common_years'?

src/cftime/_cftime.pyx
```diff
@@ -320,7 +323,10 @@ UNIT_CONVERSION_FACTORS = {
     "days": 86400 * 1000000,
     "d": 86400 * 1000000,
     "month": 30 * 86400 * 1000000,  # Only allowed for 360_day calendar
-    "months": 30 * 86400 * 1000000
+    "months": 30 * 86400 * 1000000,
+    "common_year": 365 * 86400 * 1000000, # Only allowed for 365_day and no_leap calendars
+    "years": 365 * 86400 * 1000000 # Only allowed for 365_day and no_leap calendars
```
jswhit, line 328: change 'years' to 'common_years'

andersy005: Done!

src/cftime/_cftime.pyx
andersy005: Done!

andersy005:
@jswhit, the CI isn't running because this is my first time contribution to cftime. could you approve running workflows when you get a moment?

jswhit (approved):

Issue #248, opened 2021-07-05
Comparing cftime.datetime(0, 0, 0, calendar=None)


bjlittle:
## Question
For a `cftime.datetime` where the `calendar=None`, we have a corner case where the user is able to create a valid `datetime` without explicitly specifying `has_year_zero` e.g.,
```python
>>> cftime.datetime(0, 0, 0, calendar=None)
cftime.datetime(0, 0, 0, 0, 0, 0, 0, calendar='', has_year_zero=False)
```
However, this returns a `cftime.datetime` object with a `has_year_zero=False`. Although this is correct given what the user originally specified, the `year` of the `cftime.datetime` is clearly **not** zero.

As a result, this causes issues when comparing such a `cftime.datetime` to one created with `has_year_zero=True` (see the following **Example** section).

Would it be possible to ensure that the `has_year_zero` property is verified and correct given the `year` at creation time and also if the user alters the `year` though the `replace` method i.e., keep the `has_year_zero` internally consistent given the `year` value of the instance? 

## Example
Existing behaviour:
```python
from cftime import datetime

>>> a = datetime(0, 0, 0, calendar=None)
>>> b = datetime(0, 0, 0, calendar=None, has_year_zero=True)
>>> a == a
True
>>> b == b
True
>>> a == b
TypeError: cannot compare cftime.datetime(0, 0, 0, 0, 0, 0, 0, calendar='', has_year_zero=False) and cftime.datetime(0, 0, 0, 0, 0, 0, 0, calendar='', has_year_zero=True)
```
```python
>>> dt = datetime(1970, 1, 1, calendar=None)
>>> dt
cftime.datetime(1970, 1, 1, 0, 0, 0, 0, calendar='', has_year_zero=False)
>>> dt.replace(year=0)
cftime.datetime(0, 1, 1, 0, 0, 0, 0, calendar='', has_year_zero=False)
```
Proposed behaviour:
```python
from cftime import datetime

>>> a = datetime(0, 0, 0, calendar=None)
>>> b = datetime(0, 0, 0, calendar=None, has_year_zero=True)
>>> a == a
True
>>> b == b
True
>>> a == b
True
```
```python
>>> dt = datetime(1970, 1, 1, calendar=None)
>>> dt
cftime.datetime(1970, 1, 1, 0, 0, 0, 0, calendar='', has_year_zero=False)
>>> dt.replace(year=0)
[...]

PR #249, opened 2021-07-08
address issue #248

jswhit:
This PR changes the behavior of cftime.datetime initialization.  Previously, specifying year=0 without the has_year_zero kwarg will raise an exception if the default  for that calendar is has_year_zero=False.  With this change specifying year=0 without a has_year_zero kwarg will cause has_year_zero to be set to True. 

The assumption here is that if the user is asking for an instance for year=0, they implicitly want the version of the calendar that includes year zero.

@spencerkclark would appreciate it if you would take a look at this and make sure it doesn't break xarray

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -5,6 +5,7 @@ version 1.5.1 (not yet released)
  * check consistency of year arg and has_year_zero kwarg in cftime.datetime
    (issue #248).  Also assume if has_year_zero not specified it should be True
    if year=0. Allow replace method to change has_year_zero.
+ * '360_day' was missing from list of 'idealized' calendars.
 
 version 1.5.0 (release tag v1.5.0.rel)
 ======================================
--- a/Changelog
+++ b/Changelog
@@ -3,7 +3,7 @@ version 1.5.1 (not yet released)
  * added support for "common_year" and "common_years" units for "noleap" 
    and "365_day" calendars (issue #5, PR #246)
  * check consistency of year arg and has_year_zero kwarg in cftime.datetime
-   (issue #248).  Also ssume if has_year_zero not specified it should be True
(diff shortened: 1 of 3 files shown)


jswhit:
@bjlittle and @spencerkclark - if you have a chance your feedback would be welcome.  I plan to merge later next week.

spencerkclark:
Thanks for the ping @jswhit.  I believe everything in xarray would still work with this change.  That said, from a user experience perspective, if this change is adopted we might move toward explicitly specifying our year-zero conventions anywhere where we construct dates manually in the code.  I prefer that the error message continues to come immediately from specifying a date with an inconsistent convention rather than something downstream, since in xarray we assume the CF conventions are followed.

I see @bjlittle's point regarding if a user specifies year zero then they would likely want year zero, but at the same time I do find it a little awkward that one can effectively create two different date types with the same constructor:
```
>>> import cftime
>>> a = cftime.DatetimeGregorian(0, 1, 1)
>>> b = cftime.DatetimeGregorian(1, 1, 1)
>>> b - a
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "src/cftime/_cftime.pyx", line 1398, in cftime._cftime.datetime.__sub__
TypeError: cannot compute the time difference between dates with year zero conventions
```

jswhit:
Perhaps a warning should be raised when year zero is requested and the default has_year_zero kwarg would otherwise be False, so that the user is aware they are creating an instance that doesn't conform to CF standards/

spencerkclark:
I like the idea of including a warning.  That would be a good compromise.

jswhit:
@bjlittle I'd still like to have you look this over to make sure it addresses your original issue.  I'll go ahead and merge if I don't hear from you by this Friday (Aug 13).

Issue #251, opened 2021-08-09
Calendar-specific datetime subclasses lose their identity after pickling


spencerkclark:
After pickling it appears that calendar-specific subclasses of `cftime.datetime` objects lose their identity.  This causes integration issues between downstream libraries that depend on serialization, e.g. between `xarray` and `distributed` (pydata/xarray#5686).  

Minimal working example:
```
>>> import cftime; import pickle
>>> serialized = pickle.dumps(cftime.DatetimeNoLeap(2000, 1, 1))
>>> deserialized = pickle.loads(serialized)
>>> deserialized
cftime.datetime(2000, 1, 1, 0, 0, 0, 0, calendar='noleap', has_year_zero=True)
```
I would expect `deserialized` to retain its `cftime.DatetimeNoLeap` type.  Is there a way to make sure that the type is round-tripped through pickling?

cc: @aidanheerdegen

jswhit:
The serialization happens through the `__reduce__` method

```python
    def __reduce__(self):
        """special method that allows instance to be pickled"""
        args, kwargs = self._getstate()
        return (_create_datetime, (args, kwargs))
```
I suppose we could support serialization of datetime specific instances with a bunch if if-statements here, but on the other hand I thought datetime specfic instances were deprecated anyway (since downstream libs can just check the `calendar` instance variable instead).

spencerkclark:
Thanks @jswhit -- I sort of answered my own question -- I have a solution in #252.  We have not yet fully adopted the universal `cftime.datetime` in xarray, so we still use the subclasses internally (this was motivation for the 1.4.1 release which restored those for the time being).

jswhit:
That's a very nice solution @spencerkclark

jswhit:
closed by PR #252

aidanheerdegen:
Thanks for the prompt fix.

Do you have a timeline for the next release? This bug prevents currently used code from working correctly, but we're relying on `>1.5` for the latest `xarray` release I believe.

aidanheerdegen:
Sorry to be a pain @jswhit, but this bug is affecting users, so any update on when `1.5.1` will be released?

PR #252, opened 2021-08-09
Ensure subclass type is properly roundtripped during pickling

spencerkclark:
This PR addresses the issue in #251.  It appears there was an existing test of pickle roundtripping, but it only checked that the result was equal to the input.  It seems testing for equality does not check equality of types.  

cc: @aidanheerdegen

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -2,6 +2,8 @@ version 1.5.1 (not yet released)
 ======================================
  * added support for "common_year" and "common_years" units for "noleap" 
    and "365_day" calendars (issue #5, PR #246)
+ * fixed a bug that led to subclasses losing their type identity upon
+   pickling (issue #251, PR #252).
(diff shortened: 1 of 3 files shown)


spencerkclark:
Tests pass for me locally -- it looks like an environment issue in GitHub Actions.

jswhit:
Re-running the tests now..

Can we add another test to make sure picking/unpicking still works as expected with regular `cftime.dattime` instances?

Plus, a Changelog entry?

spencerkclark:
Thanks for the review.  I added a test for `cftime.datetime` instances and a `Changelog` entry.

shoyer:
> It seems testing for equality does not check equality of types.

this probably is also worth looking into?

anyways, thanks for tracking this down!

PR #254, opened 2021-09-22
prepare for 1.5.1 release

jswhit:


jswhit2:
@spencerkclark will this change in the proleptic_gregorian defaults (to has_year_zero=T) adversely impact xarray?

spencerkclark:
The only potential issue I can think of is that it may change answers when decoding times that were originally encoded with a reference date with a negative year when year zero was not allowed.  I think that's a pretty rare occurrence, however.

jswhit2:
@spencerkclark in that case you can always use `has_year_zero=F` to get the old behavior.

Issue #256, opened 2021-11-02
Any updates needed due to CF version 1.9 is deprecating gregorian calendar?


larsbarring:
Are there any implications for cftime because of CF version 1.9 is deprecating `gregorian` calendar as synonym for the `standard`mixed Julian/Gregorian calendar? One possible example that I recently came across is name of the datatype `DatetimeGregorian`; maybe it should be `DatetimeStandard` instead?

jswhit:
PR #257 causes a `DeprecationWarning` to be raised if calendar='gregorian' is used.  A `DatetimeStandard` class was added, but `DatetimeGregorian` is kept for backwards compatibility.

davidhassell:
Hello, Is it right to raise a warning if `calendar='gregorian' is used`? For the nearly all existing CF-netCDF datasets at time of writing (i.e. those at CF<=1.8), `gregorian` is valid and so using cftime to parse such a file would, I feel, lead to an inappropriate message. I think silently changing the calendar to `standard` in such cases is fine, as this PR would do without these warnings. What do you think?

This warning pops up in two places:
https://github.com/Unidata/cftime/blob/67eeb688edc35ded0b3e20d96943eb929465f30f/src/cftime/_cftime.pyx#L1053-L1054
https://github.com/Unidata/cftime/blob/67eeb688edc35ded0b3e20d96943eb929465f30f/src/cftime/_cftime.pyx#L1770-L1771

Thanks, David

larsbarring:
I raised this issue as a question, and *think* that it might be reasonable to automagically change from`gregorian` calendar to `standard` as they were synonyms. But I am not the right person to comment whether it should be done silently or not.

jswhit:
I went ahead and removed the DeprecationWarning, and now gregorian is just silently changed to standard.  I think @davidhassell  is right, since 'gregorian' has been in use since forever, that warning would be annoying.

PR #257, opened 2021-11-05
address deprecation of calendar='gregorian' in CF v1.9 (issue #256)

jswhit:
add a DeprecationWarning if calendar='gregorian' used, replace DatetimeGregorian with DatetimeStandard (but keep DatetimeGregorian for backwards compatibility)

### DIFF
--- a/.github/workflows/miniconda.yml
+++ b/.github/workflows/miniconda.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


jswhit:
We could add a CFversion attribute to cftime.datetime and only raise the warnings if CFversion >= 1.9.

jswhit:
`DeprecationWarning` has been removed, and `cftime.datetime.__add__` and `__sub__` still return `DatetimeGregorian` objects so as not to break xarray

Issue #269, opened 2022-02-19
Potential approaches for performance improvements in `cftime.num2date`


spencerkclark:
### Issue

While debugging a user issue, a performance limitation of `cftime.num2date` was put in sharp relief in pydata/xarray#6284.  Essentially `num2date` gets slower the farther the encoded times are from their reference date.  We sort of knew about this when we adopted the new decoding algorithm in #176, but it didn't seem that serious until we encountered an array with 15+ million times.  Arrays of times that large are admittedly rare, but it could be good to explore some options for improving the performance of `num2date`, since, as suggested by @rabernat and as I show below, there is some room for improvement.

A minimal example that illustrates this performance issue is the following:
```
In [1]: import cftime; import numpy as np

In [2]: encoded_times = 365 * 2000 + np.linspace(0, 30, 10000)

In [3]: %timeit cftime.num2date(encoded_times, "days since 0001-01-01", "proleptic_gregorian")
3.21 s ± 34.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

If we bring the reference date closer to the encoded times, we obtain an over 250x speedup.

```
In [4]: encoded_times = np.linspace(0, 30, 10000)

In [5]: %timeit cftime.num2date(encoded_times, "days since 1999-09-04", "proleptic_gregorian")
11.5 ms ± 142 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

This example uses 10,000 times, and even there 3 seconds is slow.  With 15 million times, it would take over an hour.  With a better reference date, however, 15 million times would take about 15 seconds (still slow, but bearable).  

### Possible solutions

@rabernat suggested two potential performance optimizations that seem fairly natural to try:

> One possible performance optimization you could make in cftime.num2date would be to use the true reference date just for the first value of the times arrray. Then you could say times_delta = times - times[0] (re-reference everything to the first timestep) and update the reference date accordingly. Going even further, you could imagine computing each time relative as a timedelta relative to the previous timestep.

### Results

[...]

PR #270, opened 2022-02-20
Implement optimized `num2date` algorithm

spencerkclark:
This implements the "Iterate using relative timedeltas with sorting" approach described in #269, handling the zero-size and masked array edge cases as well.  It addresses the performance issue associated with using a distant reference date in encoding the times (what originally motivated this work), and also provides a nice performance boost for time arrays that span a long time range.

While other simpler approaches can be marginally faster in some circumstances, this approach is definitely the most robust to different arrangements of times (the "Unordered long" case in #269 being the most difficult for other algorithms). 

Thanks @rabernat for the clever idea!

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,6 +1,8 @@
 since version 1.5.2
 ===================
  * fix for masked array inputs (issue #267).
+ * improved performance of the num2date algorithm, in some cases providing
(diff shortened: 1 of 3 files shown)


jswhit:
Looks good @spencerkclark,thank you!  I'll merge this at the end of the week if there are no comments.

PR #278, opened 2022-06-12
Added to_tuple method

mcuntz:
I added a `to_tuple` method to `real_datetime` and made the `to_tuple` method of `datetime` accessible.

I wanted this because firstly the `timetuple` method omits the microseconds.
And secondly, using `to_tuple` allows me to get all years, for example, from an array of datetimes pretty quickly:
```
dtin = [cftime.real_datetime(1990, 1, 1), cftime.real_datetime(1991, 1, 1)]
iarrays = np.array([ dt.to_tuple() for dt in dtin ])
years  = iarrays[:, 0]
```
This was the quickest way to achieve this that I could find.

`to_tuple` should work for `cftime.datetime` objects as well. However I had to change `cdef` to `def` for the method of `cftime.datetime`. I do not code Cython, so this was a guess.

I created some tests for the `to_tuple` method.

Kind regards,
Matthias

### DIFF
--- a/src/cftime/_cftime.pyx
+++ b/src/cftime/_cftime.pyx
@@ -67,7 +67,13 @@ class real_datetime(datetime_python):
             return _dayspermonth_leap[self.month-1]
         else:
             return _dayspermonth[self.month-1]
-    def to_tuple(self):
+    def totuple(self):
+        """Turn a datetime instance into a tuple of integers. Elements go
+        in the order of decreasing significance, making it easy to compare
+        datetime instances. Parts of the state that don't affect ordering
+        are omitted. d.totuple() is equivalent to (d.year, d.month, d.day,
+        d.hour, d.minute, d.second, d.microsecond).
+        Compare to timetuple()."""
         return (self.year, self.month, self.day, self.hour, self.minute,
                 self.second, self.microsecond)
     nanosecond = 0 # workaround for pandas bug (cftime issue #77)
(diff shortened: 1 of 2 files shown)


mcuntz:
Making the documentation fails on Python 3.9, ubuntu-latest, x32 due to:
```
Warning, treated as error:
[24](https://github.com/Unidata/cftime/runs/6852257202?check_suite_focus=true#step:7:25)
Invalid configuration value found: 'language = None'. Update your configuration to a valid langauge code. Falling back to 'en' (English).
```
 This warning is also in the master branch, so I ignore it.

jswhit:
I think with this change we no longer need the to_tuple module function.  I suggest removing that function and changing

```python
return PyObject_RichCompare(dt.to_tuple(), to_tuple(other), op)
```
to
```python
return PyObject_RichCompare(dt.to_tuple(), other.to_tuple(), op)
```
Would be a good idea to copy the docstring for the to_tuple function to the to_tuple methods.

mcuntz:
Hey Jeff @jswhit ,

It feels to more like a method one would like to have rather than a module function. But you are correct that I would just need the to_tuple module function.
I added the docstrings, and renamed it to totuple for consistency because all other methods and public functions have no underscore such as toordinal. I left the totuple module function in PyObject_RichCompare because at this point, the other could be a real_datetime but also a real datetime.datetime object.

But as said, I agree that a public to_tuple module function would be sufficient. I figured that you could just do that and do not need an PR for this so I went with my suggestion. If you will go the way of the public module function, then just delete this PR.

Kind regards,
Matthias

jswhit:
I exposed the to_tuple function in the public API as part of PR #281

jswhit:
superseded by PR #281

Issue #279, opened 2022-06-24
New Numpy Version Issues


mgrover1:
To report a non-security related issue, please provide:

* Issues with version 1.6.0
* Environment Information (i.e. Operating System, compiler info, java version, python version, etc.)
  * Windows ()
  * All Python versions (3.6, 3.7, 3.8, 3.9, 3.10)
* The issue: When updating to the numpy version (1.23.0), there are issues with the num2date module, as shown below from our Py-ART CI logs. It seems to tied to the Cython num2date function.
  * [Link to log](https://github.com/ARM-DOE/pyart/runs/7038802219?check_suite_focus=true)

```bash
pyart\util\datetime_utils.py:67: in datetime_from_grid
[4680](https://github.com/ARM-DOE/pyart/runs/7038802219?check_suite_focus=true#step:8:4681)
    return num2date(grid.time['data'][0], grid.time['units'],
[4681](https://github.com/ARM-DOE/pyart/runs/7038802219?check_suite_focus=true#step:8:4682)
src\cftime\_cftime.pyx:584: in cftime._cftime.num2date
[4682](https://github.com/ARM-DOE/pyart/runs/7038802219?check_suite_focus=true#step:8:4683)
    ???
[4683](https://github.com/ARM-DOE/pyart/runs/7038802219?check_suite_focus=true#step:8:4684)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
[4684](https://github.com/ARM-DOE/pyart/runs/7038802219?check_suite_focus=true#step:8:4685)
[4685](https://github.com/ARM-DOE/pyart/runs/7038802219?check_suite_focus=true#step:8:4686)
>   ???
[4686](https://github.com/ARM-DOE/pyart/runs/7038802219?check_suite_focus=true#step:8:4687)
E   TypeError: '<' not supported between instances of 'numpy.longdouble' and 'int'
```

It only fails on Windows, and the issue only occurs with the most recent NumPy update.

jswhit:
Potential solution in PR #281

jswhit:
PR #281 merged. A 1.6.1 release is coming with this fix.

PR #280, opened 2022-06-28
add smoke test for #279

ocefpaf:
Nothing to see here (yet), just testing what is going on with #279.


ocefpaf:
We don't need the test I added. The failure is being captured in the current tests. I'll update the testing framework a bit to make it easier to see these failures and then investigate why latest numpy is causing this.

ocefpaf:
This is in #281 now. Closing it.

PR #281, opened 2022-06-28
testing solutions for issue #279

jswhit:


.github/workflows/miniconda.yml
```diff
@@ -20,6 +20,7 @@ jobs:
         exclude:
           - os: macos-latest  
             platform: x32
+        fail-fast: false
```
ocefpaf, line 23: This needs to be aligned with `matrix` on the top.

neutrinoceros:
Hi ! May I ask what would be the expected timescale for this patch to be released ? The incompatibility with numpy 1.23 is causing a serious instability in yt's CI (two levels downstream, via pyart). I'd just like to know whether it'd be worth working on a temporary workaround. Thanks !

ocefpaf:
@jswhit just to add another data point, conda-forge ppc builds were also failing with a segfault and this PR solved it too. See https://github.com/conda-forge/cftime-feedstock/pull/47

ocefpaf:
> I will merge this today and then being the process of creating a 1.6.1 release.

BTW, PR #282 should make it easier to issue releases. I'll finish it as soon as this one is merged.

PR #284, opened 2022-06-30
Roundtrip does not have microsecond precision

mcuntz:
I do not get the correct date back with date2num and num2date:
```
from cftime import datetime, date2num, num2date

testdate = (1271, 3, 18, 19, 41, 33)
calendar = 'noleap'
units = 'days since 0000-01-01 12:00:00'
date = datetime(*testdate, calendar=calendar)
num = date2num(date, units, calendar=calendar)
date2 = num2date(num, units, calendar=calendar)
print(date, date2)

```
Gives:
1271-03-18 19:41:33 1271-03-18 19:41:32.999998

I am using cftime 1.6.0 with Python 3.10.4.
I tracked it down to the line 277 in src/cftime/_cftime.pyx:
`            times.append(td / unit_timedelta)`
The Python docu says about timedelta objects:
"f = t2 / t3    Division of overall duration t2 by interval unit t3. Returns a float object."

So this returns a float object which is 64 bit and hence not enough for microsecond precision.
I hence tried:
```
tt = (np.longdouble(td.total_seconds()) /
        np.longdouble(unit_timedelta.total_seconds()))

```
But this was imprecise to 1 microseconds in some cases because to_seconds() returns also float64.
So I changed to microseconds first and then do the division:
```
tt = (np.rint(np.longdouble(td.total_seconds()) * 1000000.) /
        np.rint(np.longdouble(unit_timedelta.total_seconds()) * 1000000.))

```
This seems to work.

But I do not know what will happen on Windows, which apparently has no float128 (noted somewhere in _cftime.pyx) . I cannot test it.
I also had to change two places in the tests where the result of date2num were checked to be float64 (or compared against a float64 constant).

Kind regards,
Matthias

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -2,6 +2,9 @@ version 1.6.1 (release tag v1.6.1rel)
 =====================================
  * fix failing tests on windows with numpy 1.23.0 (issue #278)
  * expose to_tuple module function in public API.
+ * longdouble keyword in date2num so that a roundtrip from a time to a date
+   and back again does not lose microsecond precision when the units require
+   the times be encoded as floating point values (PR #284)
 
 version 1.6.0 (release tag v1.6.0rel)
 =====================================
@@ -13,18 +16,18 @@ version 1.6.0 (release tag v1.6.0rel)
 
 version 1.5.2 (release tag v1.5.2rel)
 =====================================
- * silently change calendar='gregorian' to 'standard' internally, 
+ * silently change calendar='gregorian' to 'standard' internally,
    since 'gregorian' deprecated in CF v1.9 (issue #256).
  * add "is_leap_year" function (issue #259).
  * wheels that work on Apple M1 (arm64) available on pypi.
-   
+
 version 1.5.1.1
 ===============
  * no code changes, just new binary wheels for python 3.10.
 
 version 1.5.1 (release tag v1.5.1.rel)
 ======================================
- * added support for "common_year" and "common_years" units for "noleap" 
+ * added support for "common_year" and "common_years" units for "noleap"
    and "365_day" calendars (issue #5, PR #246)
  * check consistency of year arg and has_year_zero kwarg in cftime.datetime
    (issue #248).  Also assume if has_year_zero not specified it should be True
@@ -39,7 +42,6 @@ version 1.5.1 (release tag v1.5.1.rel)
    for this calendar). Issue warning when trying to
    to create a cftime.datetime instance that is not allowed in CF (PR #238).
 
-
 version 1.5.0 (release tag v1.5.0.rel)
 ======================================
  * clean-up deprecated calendar specific subclasses (PR #231).
@@ -47,13 +49,13 @@ version 1.5.0 (release tag v1.5.0.rel)
    (via `cftime.datetime.__format__`) PR #232.
  * add support for astronomical year numbering (including year zero) for
    real-world calendars using 'has_year_zero' cftime.datetime kwarg (PR #234).
-   Default is False for 'real-world' calendars ('julian', 'gregorian'/'standard', 
+   Default is False for 'real-world' calendars ('julian', 'gregorian'/'standard',
    'proleptic_gregorian'). Ignored for idealized calendars like '360_day
    (they always have year zero).
- * add "change_calendar" cftime.datetime method to switch to another 
+ * add "change_calendar" cftime.datetime method to switch to another
    'real-world' calendar. Enable comparison of cftime.datetime instances
    with different 'real-world' calendars (using the new change_calendar method)
- * remove legacy `utime` class, and legacy `JulianDayFromDate` and 
+ * remove legacy `utime` class, and legacy `JulianDayFromDate` and
    `DateFromJulianDay` functions (replaced by `cftime.datetime.toordinal`
    and `cftime.datetime.fromordinal`).  PR #235.
  * Change ValueError to TypeError in __sub__ (issue #236, PR #236).
@@ -61,7 +63,7 @@ version 1.5.0 (release tag v1.5.0.rel)
 version 1.4.1 (release tag v1.4.1.rel)
 ======================================
  * Restore use of calendar-specific sub-classes in `cftime.num2date`,
-   `cftime.datetime.__add__`, and `cftime.datetime.__sub__`.  The use of them 
+   `cftime.datetime.__add__`, and `cftime.datetime.__sub__`.  The use of them
    will be removed in a later release.
  * add 'fromordinal' static method to create a cftime.datetime instance
    from a julian day ordinal and calendar (inverse of 'toordinal').
@@ -72,9 +74,9 @@ version 1.4.0 (release tag v1.4.0.rel)
    and times allow.  Previously this would only be true if the units were
    'microseconds' (PR #225).  In other circumstances, as before, `cftime.date2num`
    will return an array of floats.
- * Rewrite of julian day/calendar functions (_IntJulianDayToCalendar and 
+ * Rewrite of julian day/calendar functions (_IntJulianDayToCalendar and
    _IntJulianDayFromCalendar) to remove GPL'ed code.  cftime license
-   changed to MIT (to be consistent with netcdf4-python). 
+   changed to MIT (to be consistent with netcdf4-python).
  * Added datetime.toordinal() (returns julian day, kwarg 'fractional'
    can be used to include fractional day).
  * cftime.datetime no longer uses calendar-specific sub-classes.
@@ -100,7 +102,7 @@ version 1.3.0 (release tag v1.3.0rel)
    The calendar specific sub-classes are now deprecated, but remain for now
    as stubs that just instantiate the base class and override __repr__.
  * update regex in _cpdef _parse_date so reference years with more than four
-   digits can be handled. 
+   digits can be handled.
  * Change default calendar in cftime.date2num from 'standard' to None
    (calendar associated with first input datetime object is used).
  * add `cftime.datetime.tzinfo=None` for compatibility with python datetime
@@ -131,7 +133,6 @@ version 1.2.0 (release tag v1.2.0rel)
  * utime.date2num/utime.num2date now just call module level functions.
    JulianDayFromDate/DateFromJulianDay no longer used internally (PR #180).
 
-
 version 1.1.3 (release tag v1.1.3rel)
 =====================================
  * add isoformat method for compatibility with python datetime (issue #152).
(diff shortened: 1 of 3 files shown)


mcuntz:
I see that Windows is a problem. It just the float64 calculations. I don't know how to deal with it.
Any help would be appreciated.

jswhit:
those tests will have to be skipped or xfailed on windows, using something like this

```python
if  sys.platform.startswith("win"):
    pytest.skip("skipping tests that require float128 on windows")
```

spencerkclark:
@mcuntz indeed this is not surprising, given that using units of `"days since 0001-01-01 12:00:00"` require that the difference between your specified date and the reference date be represented using a floating point value.  Is there a reason you cannot use something like `units = "seconds since 0001-01-01 12:00:00"`?  In that case the difference can be represented using an integer, enabling an exact round-trip.

More generally, while it may not be the most human-readable, encoding your times with units of `"microseconds since <reference date>"` will enable universal exact round-tripping, since it will always be possible to represent differences from the reference date with integers.

jswhit:
@mcuntz I agree with @spencerkclark here, rounding to the nearest integer enables round-tripping in your case, at the expense of accuracy.  For example, if you started with

1271-03-18 19:41:33 1271-03-18 19:41:32.999998

your mods would not preserve the round trip.  Using units that can represent the time interval as an integer seems like the best way to go.

jswhit:
Consider the case where you start with 

1271-03-18 19:41:33 1271-03-18 19:41:32.999998

With your mods the round trip is

1271-03-18 19:41:33 1271-03-18 19:41:33

whereas the original code preserves the round trip.  So you're getting the round trip in the case of integer seconds at the expense of it when the the original value is not an integer second. I agree with @spencerkclark here that it's probably better to just use units that can represent the time intervals as integers.

mcuntz:
@spencerkclark and @jswhit 

I want to have the round trip working independent of the units. I am working with land surface models and they have all kinds of units (some of them are not really CF conform). So I cannot assure that the units are always the most adapted to the dates.

By the way, `"microseconds since <reference date>"` is not allowed.
```
  File "src/cftime/_cftime.pyx", line 103, in cftime._cftime._dateparse
    raise ValueError(
ValueError: units must be one of 'seconds', 'minutes', 'hours' or 'days' (or singular version of these), got 'microsecondseconds'
```

So I gave it another try which works for "1271-03-18 19:41:33" and for "1271-03-18 19:41:32.999998" with units of 'seconds', 'minutes', 'hours', and 'days'. I have just committed to look what happens on Windows.

mcuntz:
@spencerkclark

sorry, my fault, microseconds is allowed, I had a typo. So I added it into the roundtrip test.
I also added that the test is not done on windows with the lines from @jswhit .
There is a point in the tests (around line 2051) where it is assured that date2num gives float64. I have changed that to float64 on windows and float128 otherwise.

spencerkclark:
Thanks @mcuntz -- I'm liking your approach more now, since it casts directly from integer values to float128 values, rather than from float64 values to float128 values and no longer makes any rounding assumptions.  A concern I have, however, is that float128 values cannot be saved to netCDF files.  Do we still get the benefits of this approach if we cast the final result back to float64?

> sorry, my fault, microseconds is allowed, I had a typo.

No worries.  That error message can definitely be improved; it certainly makes it seem like microseconds wouldn't be allowed.  

> I want to have the round trip working independent of the units. I am working with land surface models and they have all kinds of units (some of them are not really CF conform). So I cannot assure that the units are always the most adapted to the dates.

Can you give a little more detail regarding your use case with a land surface model?  Is the issue that you load in a file produced by a land surface model, convert its time coordinate to cftime datetimes, do some transformation on other data variables in the file, and then write the resulting dataset back out to a file using the original units and find that the times are not round tripped?

jswhit:
@mcuntz thanks for updating the PR to address my concerns.  Please also add a Changelog entry, and fix up the conflicts in test_cftime.py

jswhit:
@spencerkclark asked "Do we still get the benefits of this approach if we cast the final result back to float64?" If the result of the quotient computed using longdoubles is cast to float64, many of the roundtrip tests fail.  So - I think the answer to this question is no.

mcuntz:
Dear @spencerkclark and @jswhit 

You wrote that netcdf files do not allow float128. So I am a bit confused about the microsecond precision now. Looks to me that there is only microsecond precision with the units "microseconds since ...".
After testing a bit more, I found the not surprising fact that we get great roundtrip precision with float64 as long as the number is an integer multiple of the unit, i.e. I get only microseconds back if I use "microseconds since ...", but I also only get the correct seconds back with the units "seconds since ..." if I have no microseconds, and so on with minutes, hours, etc. 

So this could be an angle of attack to say that we test this fact in date2num and if we loose precision with float64 we return float128, or simply write a warning that one looses precision with this date-units combination.

There are plenty of other possibilities such as having a keyword float128=True (default False) to have 
float128 output. Even a wrapper function date2num2 (or date2num128) that calls date2num with float128=True is imaginable. But all this probably depends on your plans.

To answer the question about my use cases of @spencerkclark : the netcdf files I get have all different kinds of calendars. The model Orchidee, for example, is using the Lilian calendar, where you get seconds precision with float32. The standard units of the climate data operators (cdo) are "day as %Y%m%d.%f". The standard output of the ECHAM climate model is decimal year such as 2022.5082191780823 for noon of today.
For a long time I had a routines that handle these dates/calendars. I am currently updating all my routines (https://github.com/mcuntz/pyjams). So I thought I built more upon cftime for non-CF calendars (https://github.com/mcuntz/pyjams/blob/main/src/pyjams/datetime.py). Unfortunately, all variables of datetime are readonly in cftime and I could not inherit from cftime. But I wanted the same logic of cftime and so I had to copy/paste a lot of code, often changing it only slightly to allow other calendars, for example. It will be very hard to keep up with improvements/bug fixes in cftime. But whenever possible I call the cftime routines.

That said, I probably abuse date2num and num2date a lot in my codes. I use it to calculate timedeltas . I check for example that I have equidistant timesteps in observational data. My pyjams.date2num can also interpret date strings (no sophisticated strptime, only standard formats). So I use np.diff on the output of date2num. Or I want to know/compare/whatever the time between two dates, I simply do `pyjams.date2num('1990-01-01 12:30") - pyjams.date2num('1990-01-15 13:30")`.

jswhit:
I still think this is a worthwhile update, even though the extra precision will be lost when the times are written to the netcdf file.  There's just no way to maintain roundtrip precision for arbitrary units and time intervals using 64-bit floats.  If users really need it, they should stick to units that that are integers multiples of the time discretization.  I don't think it's the job of cftime to check for that.

spencerkclark:
I agree this is a worthwhile update, with a preference that we default (e.g. through a keyword argument) to returning float64 data.  If we default to float128, changes will need to be made in xarray to explicitly do the casting when saving to a netCDF file (I don't believe xarray uses `date2num` anywhere else).

Thanks for those further details @mcuntz.  For computing timedeltas I tend to difference `cftime.datetime` objects directly, which is a microsecond-exact operation and returns `datetime.timedelta` objects, but I can understand why it might be more convenient to have a numeric representation of the time difference.

> The standard output of the ECHAM climate model is decimal year such as 2022.5082191780823 for noon of today.

This seems unfortunate...in these circumstances in xarray at least there are utilities for post-hoc rounding the decoded dates to a certain frequency.  As @jswhit wrote, by default, I don't think there is a way we can do any better in these instances since we are constrained by netCDF in this regard.

jswhit:
I see that HDF5 has a long double datatype (H5T_NATIVE_LDOUBLE) so in principle netcdf-c could support this.  Should we open a netcdf-c issue with a feature request?

spencerkclark:
> netcdf4-python will automatically cast to the variable type when writing (so assigning a float128 numpy array to a float64 netcdf variable is OK). @spencerkclark does this not work for xarray?

Indeed it does not.  I tried with this example with this branch of cftime and got this error message:
```
>>> import xarray as xr
>>> da = xr.DataArray([cftime.DatetimeNoLeap(2000, 1, 1, 0, 1, 1)], dims=["a"], coords=[[0]], name="foo")
>>> da.encoding["units"] = "days since 2000-01-01"
>>> da.to_dataset().to_netcdf("test.nc")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/spencer/software/xarray/xarray/core/dataset.py", line 1886, in to_netcdf
    return to_netcdf(  # type: ignore  # mypy cannot resolve the overloads:(
  File "/Users/spencer/software/xarray/xarray/backends/api.py", line 1210, in to_netcdf
    dump_to_store(
  File "/Users/spencer/software/xarray/xarray/backends/api.py", line 1257, in dump_to_store
    store.store(variables, attrs, check_encoding, writer, unlimited_dims=unlimited_dims)
  File "/Users/spencer/software/xarray/xarray/backends/common.py", line 267, in store
    self.set_variables(
  File "/Users/spencer/software/xarray/xarray/backends/common.py", line 305, in set_variables
    target, source = self.prepare_variable(
  File "/Users/spencer/software/xarray/xarray/backends/netCDF4_.py", line 490, in prepare_variable
    nc4_var = self.ds.createVariable(
  File "netCDF4/_netCDF4.pyx", line 2768, in netCDF4._netCDF4.Dataset.createVariable
  File "netCDF4/_netCDF4.pyx", line 3760, in netCDF4._netCDF4.Variable.__init__
TypeError: illegal primitive data type, must be one of dict_keys(['S1', 'i1', 'u1', 'i2', 'u2', 'i4', 'u4', 'i8', 'u8', 'f4', 'f8']), got float128
```

> I see that HDF5 has a long double datatype (H5T_NATIVE_LDOUBLE) so in principle netcdf-c could support this. Should we open a netcdf-c issue with a feature request?

I noticed this too.  I'm guessing it would take some time for an update to work through all the repos, but it seems like a reasonable request.

davidhassell:
Hi - I'd just like to highlight that the CF conventions enumerates its valid datatypes, rather than just inheriting the datatypes that are supported by netCDF-4 (https://cfconventions.org/cf-conventions/cf-conventions.html#_data_types). So if netCDF-4 were to be extended to allow long doubles, there should ideally be a PR to CF as well. This not a requirement, of course, but if you wanted to be CF-compliant ...

This is not in any way a problem, more a placeholder to remind me/us/someone to act on this if needs be.

Thanks,
David

mcuntz:
I decided for the keyword `longdouble` to date2num. I took `longdouble` instead of `float128` because you never know CPU designers will come up with ;-)

I guess this allows microsecond precision (for me ;-) and leaves time to netcdf-c and xarray. Perhaps @spencerkclark could ask xarray to automatically cast to the variable type when writing, as netcdf4-python is apparently doing, if this fits to the spirit of xarray. Otherwise we probably have to wait if netcdf-c picks up the suggestion from @jswhit and then xarray will probably allow for longdouble subsequently as well.

jswhit:
I don't think netcdf-c will support longdouble for a long time, if ever.  HDF5 apparently does not have a long double file type, just an in-memory type that maps to long double.  I think the problem is that long double support varies widely across compilers and architectures.    @mcuntz I think adding the longdouble kwarg is a good compromise for now, it's default can be true at least until xarray has the ability to cast to the variable dtype.  The only question I have now if whether it is better to just cast the result of the longdouble quotient to float64 is longdouble=False.

spencerkclark (commented):
Thanks @mcuntz!  Yes, I can work on getting casting implemented in xarray.

> The only question I have now if whether it is better to just cast the result of the longdouble quotient to float64 if longdouble=False.

I also wonder about this tradeoff.  Though I haven't done any profiling, the only potential downside I see of the casting approach is that it might be slower.  In terms of accuracy, I feel like keeping things in terms of integers and longdoubles as long as possible can only help.

mcuntz:
I just amended the `Changelog` as proposed by @spencerkclark O.K. for no longdouble in netcdf-c as explained by @jswhit  But I guess it is nice to have the precision in the Python universe so I can continue abusing cftime to do what it was never meant to do or to use it without having to think to much ;-)

I adapted the `benchmark.py` script in the `test` directory (temporarily). `date2num` was only using integer arithmetic. So I changed the units so that it had to use floating point arithmetic and ran 100,000 instead of 10,000 conversions. I either converted the longdouble calculations to float (longdouble=True) or used datetime/timedelta, i.e. `td/unit_timedelta` (longdouble=False). Results were:
longdouble=True

> processing 100000 values...
> num2date took 0.15936945599969476 seconds
> date2num took 0.6355689120027819 seconds 

longdouble=False

> processing 100000 values...
> num2date took 0.1580735219977214 seconds
> date2num took 0.35811377000209177 seconds

So passing via longdouble and casting float takes about 80% more time.

I also printed the float of the longdouble calculations and td/unit_timedelta and its difference (I transferred all tests (i.e. `test_date2num_num2date_roundtrip` and `test_num2date_precision`) to an extra script to see the output). They were exactly the same, i.e. `float(quotient)-td/unit_timedelta != 0.` was never True.

So I concluded to stay with the current logic, i.e. using td/unit_timedelta in case of longdouble=False.

jswhit:
OK thanks @mcuntz.  I will go ahead and merge now.

Issue #287, opened 2022-08-06
`cftime.num2date` fails on an empty integer-dtype array


spencerkclark:
`cftime.num2date` behaves as expected when passing in an empty array of dtype `float64`:
```
>>> cftime.num2date(np.array([], dtype="float64"), "days since 1970-01-01", calendar="proleptic_gregorian", only_use_cftime_datetimes=True)
array([], dtype=object)
```
However it raises an error if the empty array has an integer dtype:
```
>>> cftime.num2date(np.array([], dtype="int64"), "days since 1970-01-01", calendar="proleptic_gregorian", only_use_cftime_datetimes=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "src/cftime/_cftime.pyx", line 604, in cftime._cftime.num2date
  File "src/cftime/_cftime.pyx", line 442, in cftime._cftime.scale_times
  File "<__array_function__ internals>", line 180, in amin
  File "/Users/spencer/Software/miniconda3/envs/xarray-tests-py38/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 2916, in amin
    return _wrapreduction(a, np.minimum, 'min', axis, None, out,
  File "/Users/spencer/Software/miniconda3/envs/xarray-tests-py38/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 86, in _wrapreduction
    return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
ValueError: zero-size array to reduction operation minimum which has no identity
```
It would be good if both of these examples produced the same result.

cc: @dcherian
xref: pydata/xarray#6882

PR #288, opened 2022-08-06
num2date should not fail on an empty integer array

jswhit:
Closes issue #287


spencerkclark (approved):
Thanks for the quick fix @jswhit.

Issue #277, opened 2022-04-21
Implement a `strptime` method


spencerkclark:
`cftime.datetime` objects implement a `strftime` method, which one can use to convert a datetime to a string with a given format.  It would be convenient if they also implemented the inverse method, [`strptime`](https://docs.python.org/3/library/datetime.html#datetime.datetime.strptime), which allows one to parse a string with a specified format into a datetime object.  I would imagine something like this:

```
>>> cftime.datetime.strptime("2000-01-01", "%Y-%m-%d", calendar="julian")
cftime.datetime(2000, 1, 1, 0, 0, 0, 0, calendar='julian', has_year_zero=False)
```

It is kind of a handful, but [the implementation in cpython](https://github.com/python/cpython/blob/d44815cabc0a8d9932df2fa95cb374eadddb7c17/Lib/_strptime.py#L309-L557) might serve as a useful reference.  @jswhit I would be happy to help contribute this if you agree it would be a useful addition to cftime.

cc: @nbren12

nbren12:
+1 on this? Is there some reason you couldn't just use `datetime.datetime.strptime` under the hood?

spencerkclark:
> Is there some reason you couldn't just use datetime.datetime.strptime under the hood?

Different calendars.  Some cftime dates are not valid in the proleptic Gregorian calendar (e.g. February 30th in a 360 day calendar).

nbren12:
gotcha! forgive my ignorance on these things.

nbren12:
The cpython code is a bit long but not so bad. Defines a single regexp and then harvests the matched groups from it converting to int as necessary. There is some weird thread-safe caching logic to make sure re.compile isn't too slow, but that doesn't seem worth replicating.

spencerkclark:
All good, yeah, I think it would largely be copy-pastable.  I agree regarding the caching logic.

jswhit:
@spencerkclark I think this would be a great feature to have

spencerkclark:
Awesome, thanks @jswhit for confirming.  Indeed I saw your thumbs up earlier, but have just been caught up with other things lately.  Hopefully I'll have a chance to return to this soon.

jswhit:
or we could use `datetime.datetime.strptime` and then convert the result from the `proleptic_gregorian` calendar using `cftime.datetime.change_calendar`.

jswhit:
actually, `change_calendar` is not even needed.  Any reason something like this wouldn't work?

```python
[...]

PR #289, opened 2022-08-21
add a strptime static method (issue #277)

jswhit:
uses python's datetime.datetime.strptime if possible (for dates in proleptic_gregorian).
for dates that don't work with python's strptime, use a stripped-down version of _strptime.py from C-python that doesn't handle timezones and locales.

TODO: formatting for negative years 

@spencerkclark apologies if you were already working on this - I had some free time this weekend so decided to take a crack at it.

### DIFF
--- a/src/cftime/_cftime.pyx
+++ b/src/cftime/_cftime.pyx
@@ -14,6 +14,7 @@ from datetime import datetime as datetime_python
 from datetime import timedelta, MINYEAR, MAXYEAR
 import time                     # strftime
 import warnings
+from ._strptime import _strptime
 
 microsec_units = ['microseconds','microsecond', 'microsec', 'microsecs']
 millisec_units = ['milliseconds', 'millisecond', 'millisec', 'millisecs', 'msec', 'msecs', 'ms']
@@ -1245,16 +1246,23 @@ The default format of the string produced by strftime is controlled by self.form
         For a complete list of formatting directives, see section
         'strftime() and strptime() Behavior' in the base Python documentation.
         """
-        # use python's datetime.strptime to get a python datetime instance
-        # (using proleptic_gregorian calendar)
-        pydatetime = datetime_python.strptime(datestring, format)
-        # remove time zone offset
-        if getattr(pydatetime, 'tzinfo',None) is not None:
-            pydatetime = pydatetime.replace(tzinfo=None) - pydatetime.utcoffset()
-        # convert the cftime datetime instance
-        return datetime(pydatetime.year, pydatetime.month, pydatetime.day,
-                        pydatetime.hour, pydatetime.minute, pydatetime.second,
-                        pydatetime.microsecond, calendar=calendar, has_year_zero=has_year_zero)
+        # if possible use python's datetime.strptime to get a python datetime instance
+        # (works for dates in proleptic_gregorian calendar)
+        try:
+            pydatetime = datetime_python.strptime(datestring, format)
+            # remove time zone offset
+            if getattr(pydatetime, 'tzinfo',None) is not None:
+                 pydatetime = pydatetime.replace(tzinfo=None) - pydatetime.utcoffset()
+            # convert the cftime datetime instance
+            return datetime(pydatetime.year, pydatetime.month, pydatetime.day,
+                            pydatetime.hour, pydatetime.minute, pydatetime.second,
+                            pydatetime.microsecond, calendar=calendar, has_year_zero=has_year_zero)
+        # otherwise use a stripped-down version of C-python's _strptime.py
+        # (doesn't understand all possible formats, including locales and time zones)
+        except ValueError:
+            year,month,day,hour,minute,second,microsecond = _strptime(datestring,format)
+            return datetime(year,month,day,hour,minute,second,microsecond,
+                            calendar=calendar,has_year_zero=has_year_zero)
 
     def __format__(self, format):
         # the string format "{t_obj}".format(t_obj=t_obj)
--- a/src/cftime/_cftime.pyx
+++ b/src/cftime/_cftime.pyx
@@ -38,7 +38,7 @@ cdef int[12] _dayspermonth_leap = [31, 29, 31, 30, 31, 30, 31, 31, 30, 31, 30, 3
 cdef int[13] _cumdayspermonth = [0, 31, 59, 90, 120, 151, 181, 212, 243, 273, 304, 334, 365]
 cdef int[13] _cumdayspermonth_leap = [0, 31, 60, 91, 121, 152, 182, 213, 244, 274, 305, 335, 366]
 
-__version__ = '1.6.1'
+__version__ = '1.6.2'
 
 # Adapted from http://delete.me.uk/2005/03/iso8601.html
 # Note: This regex ensures that all ISO8601 timezone formats are accepted - but, due to legacy support for other timestrings, not all incorrect formats can be rejected.
--- a/src/cftime/_cftime.pyx
+++ b/src/cftime/_cftime.pyx
@@ -225,7 +225,7 @@ def date2num(dates, units, calendar=None, has_year_zero=None, longdouble=False):
             has_year_zero = _year_zero_defaults(calendar)
 
     # if calendar is None or '', use calendar of first input cftime.datetime instances.
-    # if inputs are 'real' python datetime instances, use propleptic gregorian.
+    # if inputs are 'real' python datetime instances, use proleptic gregorian.
     if not calendar:
         if all_python_datetimes:
             calendar = 'proleptic_gregorian'
@@ -695,7 +695,7 @@ def date2index(dates, nctime, calendar=None, select='exact', has_year_zero=None)
             has_year_zero = _year_zero_defaults(calendar)
 
     # if calendar is None or '', use calendar of first input cftime.datetime instances.
-    # if inputs are 'real' python datetime instances, use propleptic gregorian.
+    # if inputs are 'real' python datetime instances, use proleptic gregorian.
     if not calendar:
         d0 = dates_test.item(0)
         if isinstance(d0,datetime_python):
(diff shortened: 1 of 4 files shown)


spencerkclark:
No worries @jswhit!  I had not had a chance to start on this yet, so I am all good with you taking it on.  This will be a nice addition to cftime.  I'm happy to take a closer look at it when you are ready.

jswhit:
@spencerkclark  this is now ready for review

src/cftime/_strptime.py
```diff
@@ -0,0 +1,135 @@
...
+# DO NOT modify _TimeRE_cache or _regex_cache without acquiring the cache lock
+# first!
+_TimeRE_cache = TimeRE()
+_CACHE_MAX_SIZE = 5 # Max number of regexes stored in _regex_cache
+_regex_cache = {}
+
+def _strptime(data_string, format="%a %b %d %H:%M:%S %Y"):
+    """Return a 2-tuple consisting of a time struct and an int containing
+    the number of microseconds based on the input string and the
+    format string."""
```
spencerkclark, line 94:
```suggestion
    """Return a 7-tuple consisting of the data required to construct a datetime
    based on the input string and the format string."""
```

src/cftime/_cftime.pyx
```diff
@@ -1237,6 +1236,32 @@ The default format of the string produced by strftime is controlled by self.form
             format = self.format
         return _strftime(self, format)
 
+    @staticmethod
+    def strptime(datestring, format, calendar='standard', has_year_zero=None):
+        """
+        Return a datetime corresponding to date_string, parsed according to format, 
+        with a specified calendar and year zero convention. 
+        For a complete list of formatting directives, see section
+        'strftime() and strptime() Behavior' in the base Python documentation.
```
spencerkclark, line 1245: It might be good to document the subset of directives that are guaranteed to be supported for all calendars through the fallback internal implementation of `strptime`.

src/cftime/_cftime.pyx
```diff
@@ -1237,6 +1236,32 @@ The default format of the string produced by strftime is controlled by self.form
...
+            pydatetime = datetime_python.strptime(datestring, format)
+            # remove time zone offset
+            if getattr(pydatetime, 'tzinfo',None) is not None:
+                 pydatetime = pydatetime.replace(tzinfo=None) - pydatetime.utcoffset()
+            # convert the cftime datetime instance
+            return datetime(pydatetime.year, pydatetime.month, pydatetime.day,
+                            pydatetime.hour, pydatetime.minute, pydatetime.second,
+                            pydatetime.microsecond, calendar=calendar, has_year_zero=has_year_zero)
+        # otherwise use a stripped-down version of C-python's _strptime.py
+        # (doesn't understand all possible formats, including locales and time zones)
```
spencerkclark, line 1259: It will support the locale's month names and abbreviations, though, correct?  I think since you use the standard library `calendar` module for those [this will be the case](https://docs.python.org/3/library/calendar.html#calendar.month_name).

test/test_cftime.py
```diff
@@ -1705,6 +1705,20 @@ def test_string_format2():
...
+             "%Y-%m-%d",calendar='360_day',has_year_zero=True)
+    assert(repr(d) == "cftime.datetime(0, 2, 30, 0, 0, 0, 0, calendar='360_day', has_year_zero=True)")
+    d = cftime.datetime.strptime('-99999-02-29 10:18:32.926',\
+             '%Y-%m-%d %H:%M:%S.%f',calendar='366_day')
+    assert(repr(d) == "cftime.datetime(-99999, 2, 29, 10, 18, 32, 926000, calendar='all_leap', has_year_zero=True)")
+    d = cftime.datetime.strptime('24/Aug/-4712:17:57:26', '%d/%b/%Y:%H:%M:%S',calendar='julian')
+    assert(repr(d) == "cftime.datetime(-4712, 8, 24, 17, 57, 26, 0, calendar='julian', has_year_zero=False)")
+    d = cftime.datetime.strptime('24/August/-4712:17:57:26', '%d/%B/%Y:%H:%M:%S',calendar='julian')
+    assert(repr(d) == "cftime.datetime(-4712, 8, 24, 17, 57, 26, 0, calendar='julian', has_year_zero=False)")
+
```
spencerkclark, line 1721: Maybe add a test case that the appropriate defaults are added when only a year is provided, e.g. for `cftime.datetime.strptime("-4712", "%Y", calendar="julian")`?

spencerkclark (commented):
Awesome, thanks @jswhit, particularly for making sure that we have a fallback for some of the edge cases where the standard library `strptime` will not do.  I took a look this morning and just had a few minor comments.

src/cftime/_cftime.pyx
```diff
@@ -1237,6 +1238,32 @@ The default format of the string produced by strftime is controlled by self.form
...
+        """
+        Return a datetime corresponding to date_string, parsed according to format, 
+        with a specified calendar and year zero convention. 
+        For a complete list of formatting directives, see section
+        'strftime() and strptime() Behavior' in the base Python documentation.
+        """
+        # if possible use python's datetime.strptime to get a python datetime instance
+        # (works for dates in proleptic_gregorian calendar)
+        try:
+            pydatetime = datetime_python.strptime(datestring, format)
```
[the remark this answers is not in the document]
jswhit: good idea, if the parsing fails on one of the format directives allowed in python (for example 'G'), it will  now say 
```
'G' directive not supported for dates not valid in python proleptic_gregorian calendar
```

src/cftime/_strptime.py
```diff
@@ -0,0 +1,135 @@
...
+            year = int(found_dict['Y'])
+        elif group_key == 'm':
+            month = int(found_dict['m'])
+        elif group_key == 'd':
+            day = int(found_dict['d'])
+        elif group_key == 'H':
+            hour = int(found_dict['H'])
+        elif group_key == 'I':
+            hour = int(found_dict['I'])
+            ampm = found_dict.get('p', '').lower()
```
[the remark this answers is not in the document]
jswhit: I elected to remove the 'I' directive, so we don't have to support 'p'

src/cftime/_strptime.py
```diff
@@ -0,0 +1,135 @@
...
+        return re_compile(self.pattern(format), IGNORECASE)
+
+_cache_lock = _thread_allocate_lock()
+# DO NOT modify _TimeRE_cache or _regex_cache without acquiring the cache lock
+# first!
+_TimeRE_cache = TimeRE()
+_CACHE_MAX_SIZE = 5 # Max number of regexes stored in _regex_cache
+_regex_cache = {}
+
+def _strptime(data_string, format="%a %b %d %H:%M:%S %Y"):
```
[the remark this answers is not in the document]
jswhit: done

src/cftime/_strptime.py
jswhit: done

src/cftime/_cftime.pyx
jswhit: now reads
```python
        """
        Return a datetime corresponding to date_string, parsed according to format,
        with a specified calendar and year zero convention.
        The format directives 'y','Y','m','B','b','d','H','M','S' and 'f'
        are supported for all calendars and dates.  If the date is valid
        in the python 'proleptic_gregorian' calendar, then python's
        datetime.strptime is used. For a complete list of formatting directives
        supported in python's datetime.strptime, see section
        'strftime() and strptime() Behavior' in the base Python documentation.
        """
```

src/cftime/_cftime.pyx
jswhit: yes, you are right.  I changed this to 
```
        # (doesn't understand all possible formats, just
        # 'y','Y','m','B','b','d','H','M','S' and 'f')
```

test/test_cftime.py
jswhit: done

src/cftime/_cftime.pyx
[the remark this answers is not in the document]
spencerkclark: Thanks for adding that in the internal implementation.  I think that's a useful addition too.   My primary concern here actually relates to "false positives," i.e. cases where the standard library `strptime` succeeds, but produces a date that isn't accurate in the chosen calendar.  

E.g. something like this would succeed:
```
>>> cftime.datetime.strptime("2000-45-3", "%G-%V-%u", calendar="noleap")
```
but would return the same result (just with a different date type) as:
```
>>> cftime.datetime.strptime("2000-45-3", "%G-%V-%u", calendar="proleptic_gregorian")
```
This corresponds to a date in ISO 8601 year 2000, at ISO 8601 week 45, on a Wednesday.  However, these dates are not guaranteed to be the same in both calendars.  So I was thinking we may also want to add an error message up front for the directives I listed above for non-proleptic-Gregorian calendars.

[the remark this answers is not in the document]
jswhit: ahh, I see.  We have the `_can_use_python_datetime` function that checks to see if the instance is compatible with the python 'proleptic_gregorian' calendar.  Maybe this can be used to decide if an error should be raised.   Basically,  if the format directive is one of the excluded ones and the following is not True

```
    (calendar == 'proleptic_gregorian' and date.year >= MINYEAR and date.year <= MAXYEAR) or \
     (calendar in ['gregorian','standard'] and date > gregorian and date.year <= MAXYEAR) 
```
then raise an exception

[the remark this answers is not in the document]
jswhit: I attempted to implement this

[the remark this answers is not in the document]
spencerkclark: Perfect, yes, this is what I had in mind.  Indeed some Gregorian calendar dates will be valid too, so thanks for adding that extra logic.

src/cftime/_cftime.pyx
spencerkclark: Great, thanks.  Looks good to me.

spencerkclark (approved):
Thanks @jswhit!  This looks ready to go from my perspective now.

PR #290, opened 2022-09-12
use cibuildwheel to build all wheels

ocefpaf:
This is an attempt to unify the wheel building with tagged release. Cibuildwheel is probably the best tech at the moment to build wheels and everything (hopefully) should "just work" here. If this is acceptable I'll try netcdf4-python later.

This one builds wheels for macOS intel, M1, Windows, and Linux for Pythons 3.8, 3.9, 3.10 and 3.11.

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -1,0 +1,88 @@
+name: Wheels
+
+on:
+  pull_request:
+
+  push:
+    tags:
+      - "v*"
+
+jobs:
+  build_bdist:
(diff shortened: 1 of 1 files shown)


ocefpaf:
Things to note that are different from the current wheel building setup:

1. the Linux wheels are quite bigger here, not sure why :-/
2. this would publish separate wheels on macOS for Intel and M1 instead of the universal2 wheel

jswhit:
This looks great @ocefpaf!  If this is ready to merge I will go ahead.

ocefpaf:
> This looks great @ocefpaf! If this is ready to merge I will go ahead.

It should be ready to go. The upload to PyPI also automated and should happen with a new tag.

Issue #301, opened 2022-12-12
Unexpected results when not using separators in format code with cftime.datetime.strptime


mo-jareddrayton:
I think there may be some unexpected/incorrect behaviour with the new `cftime.datetime.strptime` method added in `1.6.2` when using format strings that don't use separators. 

Also checked with Python `3.9`
```python
import cftime
import sys
print(cftime.__version__)
print(sys.version)
```

    1.6.2
    3.11.0 | packaged by conda-forge | (main, Oct 25 2022, 06:24:40) [GCC 10.4.0]

```python
print("Case 1")
print(cftime.datetime.strptime("20200230", "%Y%m%d", "360_day"), "- Wrong year is returned for this valid '360_day' date.")
print("Case 2")
print(cftime.datetime.strptime("20200229", "%Y%m%d", "365_day"), "- Accepts an invalid '365_day' date and returns wrong year.")
print("Case 3")
print(cftime.datetime.strptime("20210229", "%Y%m%d", "proleptic_gregorian"), "- Accepts an invalid 'proleptic_gregorian' date and seemingly passes this to the non standard calendar functionality and returns wrong year.")
```

    Case 1
    20200-02-30 00:00:00 - Wrong year is returned for this valid '360_day' date.
    Case 2
    202002-02-09 00:00:00 - Accepts an invalid '365_day' date and returns wrong year.
    Case 3
    202102-02-09 00:00:00 - Accepts an invalid 'proleptic_gregorian' date and seemingly passes this to the non standard calendar functionality and returns wrong year.

jswhit:
This results from my attempt to allow for negative and > 4 digit years in the regex.  The year is parsed correctly if I revert to the original regex
```
           'Y': r"(?P<Y>\d\d\d\d)"
```
instead of
```
'Y': r"(?P<Y>[+-]?[0-9]+)"
```
which doesn't work without a separator. I really need the help of a regex guru to fix this so it works for negative years and years with more than 4 digits.

jswhit:
```'Y': r"(?P<Y>[+-]?\d\d\d\d)"``` works with 4 digit years, positive or negative, and no separators.  Looks like we have to choose between allowing greater than 4 digit years and date strings without separators.

jswhit:
i guess another way to put it is: "should we support date strings without separators?"  Doing so basically requires restricting to 4 digit years.

jswhit:
@spencerkclark

jswhit:
[...]

PR #302, opened 2022-12-15
support formats without separators in strptime

jswhit:
but no longer support years with > 4 digits.
Closes issue #301


spencerkclark (approved):
Thanks @jswhit — this looks good to me.

Issue #271, opened 2022-02-27
Test failure with Cython 3 TypeError: unsupported operand type(s) for -: 'datetime.datetime' and 'cftime._cftime.DatetimeGregorian'


opoplawski:
See https://bugzilla.redhat.com/show_bug.cgi?id=2058169

Version 1.5.2 on Fedora rawhide with Python 3.11.0a5:
```
=================================== FAILURES ===================================
________________________ TestDate2index.test_select_nc _________________________

self = <test.test_cftime.TestDate2index testMethod=test_select_nc>

    def test_select_nc(self):
        nutime = self.time_vars['time']
    
        # these are python datetimes ('proleptic_gregorian' calendar).
        dates = [datetime(1950, 1, 2, 6), datetime(
            1950, 1, 3), datetime(1950, 1, 3, 18)]
    
        t = date2index(dates, nutime, select='before')
        assert_equal(t, [1, 2, 2])
    
        t = date2index(dates, nutime, select='after')
        assert_equal(t, [2, 2, 3])
    
        t = date2index(dates, nutime, select='nearest')
        assert_equal(t, [1, 2, 3])
    
        # Test dates outside the support with select
        t = date2index(datetime(1949, 12, 1), nutime, select='nearest')
        assert_equal(t, 0)
    
        t = date2index(datetime(1978, 1, 1), nutime, select='nearest')
        assert_equal(t, 365)
    
        # Test dates outside the support with before
        self.assertRaises(
            ValueError, date2index, datetime(1949, 12, 1), nutime, select='before')
    
        t = date2index(datetime(1978, 1, 1), nutime, select='before')
        assert_equal(t, 365)
    
        # Test dates outside the support with after
        t = date2index(datetime(1949, 12, 1), nutime, select='after')
        assert_equal(t, 0)
    
        self.assertRaises(
            ValueError, date2index, datetime(1978, 1, 1), nutime, select='after')
        # test microsecond and millisecond units
        unix_epoch = "milliseconds since 1970-01-01T00:00:00Z"
        d = datetime(2038, 1, 19, 3, 14, 7)
        millisecs = int(
            date2num(d, unix_epoch, calendar='proleptic_gregorian'))
        assert_equal(millisecs, (2 ** 32 / 2 - 1) * 1000)
        unix_epoch = "microseconds since 1970-01-01T00:00:00Z"
        microsecs = int(date2num(d, unix_epoch))
[...]

PR #305, opened 2023-07-29
Set `c_api_binop_methods` Cython compiler directive to `True`

spencerkclark:
The change described [here](https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#arithmetic-special-methods) appears to be the cause of the errors reported in #271.  The documentation notes that one way to address this is to restore the old arithmetic operator behavior by setting the `c_api_binop_methods` compiler directive to `True`.  This PR makes this change.  I've confirmed locally that it addresses the issue.  

Note that this directive requires Cython >= 0.29.20, [which was released on 2020-06-10](https://cython.readthedocs.io/en/latest/src/changes.html?highlight=c_api_binop_methods#id113); I updated the `pyproject.toml` and `requirements-dev.txt` files accordingly.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -2,6 +2,8 @@ since version 1.6.2 release
 ===========================
  * add support for formats without separators in strptime (e.g. '20200229', issue #301).
    This required removing support for > 4 digit years.
+ * set the c_api_binop_methods compiler directive to True to retain Cython 0.x
+   behavior for arithmetic operators for Cython >= 3.0.0 (issue #271).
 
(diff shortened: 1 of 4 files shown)


jswhit:
thanks for tracking this down @spencerkclark !

spencerkclark:
For sure.  Thanks for the quick review and merge!

ocefpaf:
@jswhit latest release [was from Sep 2022](https://github.com/Unidata/cftime/releases/tag/v1.6.2rel), it would be nice to get a new one with this fix.

PR #306, opened 2023-07-31
Update `intersphinx_mapping` format; use `c_api_binop_methods` during coverage installs

spencerkclark:
The following deprecation warning led [the documentation build](https://github.com/Unidata/cftime/actions/runs/5701592450/job/15452452061) to fail after merging #305:

```
Warning, treated as error:
Failed to read intersphinx_mapping[https://docs.python.org/], ignored: SphinxWarning('The pre-Sphinx 1.0 \'intersphinx_mapping\' format is deprecated and will be removed in Sphinx 8. Update to the current format as described in the documentation. Hint: "intersphinx_mapping = {\'<name>\': (\'[https://docs.python.org/\](https://docs.python.org//)', None)}".https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#confval-intersphinx_mapping')
```

This PR follows the first example listed [here](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#confval-intersphinx_mapping) to address this warning, allowing the documentation build to succeed.  Sphinx 1.0 was released in 2010, so I think this updated format has been supported for a long time.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -162,4 +162,4 @@ texinfo_documents = [
 # -- Options for intersphinx extension ---------------------------------------
 
 # Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {'https://docs.python.org/': None}
+intersphinx_mapping = {'python': ('https://docs.python.org/3', None)}
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,12 @@ COMPILER_DIRECTIVES = {
     # directive was added in Cython version 0.29.20.
     "c_api_binop_methods": True
 }
+COVERAGE_COMPILER_DIRECTIVES = {
+    "linetrace": True,
+    "warn.maybe_uninitialized": False,
+    "warn.unreachable": False,
+    "warn.unused": False,
+}
 DEFINE_MACROS = [("NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION")] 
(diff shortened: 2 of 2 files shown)


spencerkclark:
As an aside, on the topic of CI, I noticed that I neglected to add the `c_api_binop_methods` compiler directive in the case that code coverage was being tracked: https://github.com/Unidata/cftime/blob/144da536e4c5eb2fea0ee44e2e2a565bfaf17cc3/setup.py#L82-L92  Should add that in a new PR?  If I recall correctly, however, we stopped running tests in that mode since it slowed the test suite down considerably.

jswhit:
> As an aside, on the topic of CI, I noticed that I neglected to add the `c_api_binop_methods` compiler directive in the case that code coverage was being tracked:
> 
> https://github.com/Unidata/cftime/blob/144da536e4c5eb2fea0ee44e2e2a565bfaf17cc3/setup.py#L82-L92
> 
> Should add that in a new PR? If I recall correctly, however, we stopped running tests in that mode since it slowed the test suite down considerably.

I think it would be fine to add it to this PR

spencerkclark:
Sounds good.  I tried to modify things such that we do not need to define the value for the `c_api_binop_methods` directive twice.  Let me know if that looks OK.

jswhit:
Looks good @spencerkclark, thank you

PR #309, opened 2023-08-15
Update cibuildwheel to 2.15.0

richli:
The motivation for this update is to start building wheels for Python 3.12 by default.

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -29,7 +29,7 @@ jobs:
           arch: AMD64
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       with:
         fetch-depth: 0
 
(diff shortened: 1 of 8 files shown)


richli:
[Changelog for 2.15.0](https://github.com/pypa/cibuildwheel/blob/v2.15.0/docs/changelog.md)

richli:
Oh, maybe this was jumping the gun a bit, numpy wheels for Python 3.12 aren't out yet, but it looks numpy 1.26 will address that.

richli:
Based on [this comment](https://github.com/numpy/numpy/issues/23808#issuecomment-1678677094) I updated the build-time and run-time dependencies for Python 3.12 so that it can use the pre-release numpy version `1.26.0b1`.

The `cftime` wheel now builds for Python 3.12, but during the test phase when installing from `requirements-dev.txt`, it cannot yet install `twine` so I've disabled that. Presumably this is only temporary and that requirement specifier can be removed later.

richli:
It turns out that reusing `requirements.txt` and `requirements-dev.txt` to specify a conda environment doesn't work with the [requirement specifiers](https://pip.pypa.io/en/latest/reference/requirement-specifiers/) that were added. So I updated the CI config to not use them.

In the process, I updated the `mamba-org/provision-with-micromamba` action, which was deprecated, to its replacement, `mamba-org/setup-micromamba`, and updated a couple of things to bring the CI up-to-date. So hopefully everything works again.

PR #310, opened 2023-10-16
py37 entered EOL

ocefpaf:
I also want to check if we can build 3.12 here.


ocefpaf:
@jswhit wheels for Python 3.12 are building just fine, see

Screenshot from 2023-10-16 13-11-50


it would be nice to mint a new release to get these published.

richli:
Should `setup_requires` and `classifiers` in `setup.py` also be updated to remove Python 3.7 support?

richli:
Also, if [NEP 29 (recommended Python versions for the numpy community)](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule) were followed, Python 3.8 support should be dropped as well.

Issue #312, opened 2023-10-20
Missing 1.6.3 sdist on pypi


mgorny:
The 1.6.3 release features only wheels on PyPI. Please upload the sdist as well for repackaging needs.

PR #313, opened 2023-10-20
build sdist too

ocefpaf:
This won't fix the current missing sdist reported in #312, we should do a manual upload of that one, but it will prevent from happening again the next time a release is minted.


jswhit:
sdist manually uploaded for 1.6.3

Issue #325, opened 2024-04-22
BUG: Not NumPy 2.0 compatible


larsoner:
Modules need to be rebuilt and released now that NumPy 2.0 is out. Importing current `cftime` from PyPI I get on Linux:
```
>>> import cftime

A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.1.0.dev0 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled with NumPy 2.0.
Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.

If you are a user of the module, the easiest solution will be to
downgrade to 'numpy<2' or try to upgrade the affected module.
We expect that some modules will need time to support NumPy 2.

Traceback (most recent call last):  File "<stdin>", line 1, in <module>
  File "/home/larsoner/python/virtualenvs/base/lib/python3.11/site-packages/cftime/__init__.py", line 1, in <module>
    from ._cftime import (datetime, real_datetime,
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/larsoner/python/virtualenvs/base/lib/python3.11/site-packages/cftime/__init__.py", line 1, in <module>
    from ._cftime import (datetime, real_datetime,
  File "src/cftime/_cftime.pyx", line 1, in init cftime._cftime
ImportError: numpy.core.multiarray failed to import (auto-generated because you didn't call 'numpy.import_array()' after cimporting numpy; use '<void>numpy._import_array' to disable if you are certain you don't need it).
```
See also https://github.com/numpy/numpy/issues/24300#issuecomment-2030603395

ocefpaf:
Numpy 2.0 is not out! There is a release candidate version for testing but the stable one isn't published yet. See https://pypi.org/project/numpy/#history.

larsoner:
Correct it's not out but the most recent advice from the linked thread is:

> With numpy 2.0.0rc1 available, everyone can start doing their 2.0-compatible releases!

ocefpaf:
Sure. Do you want to send a PR to add that to our testing?

larsoner:
> Sure. Do you want to send a PR to add that to our testing?

Looks like you do already test against 2.0:

https://github.com/Unidata/cftime/blob/d549c09638c627f6fcec3d34b0ae38c92056667c/.github/workflows/tests_latest.yml#L27-L30

And your pyproject.toml is nearly set up to build against NumPy 2.0:

[...]

PR #326, opened 2024-04-22
MAINT: Build wheels against NumPy 2.0

larsoner:
Opening just for completeness in case 2.0-compatible wheels want to be cut before 2.0 actually comes out. Probably no point in merging if new wheels won't be released before NumPy 2.0.0 is released. So feel free to close!

Related to #325

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -46,6 +46,7 @@ jobs:
           python -c "import cftime; print(f'cftime v{cftime.__version__}')" &&
           python -m pip install check-manifest cython pytest pytest-cov &&
           python -m pytest -vv {package}/test
+        CIBW_BUILD_FRONTEND: 'pip; args: --pre --extra-index-url "https://pypi.anaconda.org/scientific-python-nightly-wheels/simple"'
 
     - uses: actions/upload-artifact@v3
       with:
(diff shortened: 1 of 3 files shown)


snowman2:
This PR is not necessary with this pin https://github.com/Unidata/cftime/blob/d549c09638c627f6fcec3d34b0ae38c92056667c/pyproject.toml#L7

Added in #319

.github/workflows/cibuildwheel.yml
```diff
@@ -46,6 +46,7 @@ jobs:
           python -c "import cftime; print(f'cftime v{cftime.__version__}')" &&
           python -m pip install check-manifest cython pytest pytest-cov &&
           python -m pytest -vv {package}/test
+        CIBW_BUILD_FRONTEND: 'pip; args: --pre --extra-index-url "https://pypi.anaconda.org/scientific-python-nightly-wheels/simple"'
```
larsoner, line 49: This part (or something like it) is necessary if you want to release before 2.0 actually lands

larsoner:
Closing since it seems like this won't be useful given the release plan!

.github/workflows/cibuildwheel.yml
snowman2: https://pypi.org/project/numpy/2.0.0rc1/#files

larsoner: Ohhh yes you're right I forgot it's on PyPI proper now :facepalm:

snowman2: Thank you for being proactive and for helping with this. I am also looking forward to a numpy 2 compatible release. Just trying to save you time.

PR #334, opened 2024-06-02
ENH: Build aarch64 linux and simplify matrix

### LINKED ISSUE #333: No `aarch64` wheels on PyPI
Hello, following #243, I expected to find `aarch64` wheels on PyPI. However, at least for the latest release, there are none. Is this intentional? Could you please provide binaries for `aarch64`? Thank you!

larsoner:
This has some opinionated optional changes, let me know if they're worth keeping:

1. Instead of making a 3x3 build matrix that results in 9 entries then exclude 5 of them... just `include:` the desired ones explicitly. It seems more readable in the end to me at least!
2. Build and test wheels on PRs against `master`. Assuming it's fast enough, might as well if it's quick (but can be reverted if it's not!) since it tests a bunch of archs/oses.
3. Don't exclude testing on arm64, it's on native hardware with `macos-14` so should be fast enough hopefully?
4. Build the x86_64 macOS wheels on macos-13 which is natively Intel / x86_64 rather than cross-compiling on macos-14 (which is arm64)

And it also includes the change that I actually set out to implement, namely adding Linux `aarch64` wheels.

Closes #333

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -40,10 +40,11 @@ jobs:
         platforms: all
       if: runner.os == 'Linux' && matrix.arch == 'aarch64'
 
-
     - name: Build just oldest and newest on PRs, all on tags
       shell: bash
-      # Always omit musl 3.8 b/c NumPy does not provide wheels for it
+      # - Always omit musl 3.8 b/c NumPy does not provide wheels for it
+      # - Omit musllinux on PRs (slow)
+      # - Omit musllinux_aarch64 because it's slow and niche
       run: |
         if [[ "${{ github.event_name }}" == "pull_request" ]]; then
           CIBW_SKIP="pp* cp36-* cp37-* cp38-musllinux* cp39-* cp310-* cp311-* *musllinux*"
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -2,6 +2,8 @@ name: Wheels
 
 on:
   pull_request:
+    branches:
+      - master
 
   push:
     tags:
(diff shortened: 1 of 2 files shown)


larsoner:
... if the optional changes above are desired, I'll open a separate PR first to just make `cibuildwheel` run on PRs for example. That way you won't have to approve every commit I push here, usually messing with CIs requires some iteration and that will really slow things down, so getting a tiny PR accepted first can help!

ocefpaf:
macos-13 is being deprecated, no? I believe that cross-compiling will be the only option moving forward.

ocefpaf:
> Maybe you're thinking of macos-11 [which is indeed deprecated](https://github.blog/changelog/2024-05-20-actions-upcoming-changes-to-github-hosted-macos-runners/#macos-11-deprecation-and-removal)? I have to imagine `-13`, being the last Intel runner, will be available for some years still.

Maybe is was Azure. Too many CIs to manage.

jswhit:
ubuntu-22.04 aarch64 wheel build test is failing.

larsoner:
Okay I think this one should be good to go. Comments are in the YAML, but the short version in that I now have the wheel building run in PRs but **for PRs** cut down on which ones get built for speed. For Ubuntu, macOS, and Windows, the longest is 4m (Windows). The new `aarch64` build+test takes 14m. If this is too slow, you could cut out testing (getting it down to ~7 min) or even building them at all for PRs. On tags, all relevant wheels will be built.

jswhit:
@larsoner this looks great, thank you.  Can you add an item to the Changelog before I merge?

larsoner:
Done!

hoechenberger:
Thanks!

Issue #365, opened 2025-06-25
__richcmp__ can be very slow for larger data


ocefpaf:
To report a non-security related issue, please provide:

* the version of the software with which you are encountering an issue

cftime 1.6.4
 
* environmental information (i.e. Operating System, compiler info, java version, python version, etc.)

Linux
 
* a description of the issue with the steps needed to reproduce it

The `<` comparison below can take more than 1 min. I'm may be doing something wrong, but (if not) I wonder how we can improve the rich comparison [1] in `cftime`.

```python
In [1]: import cftime
   ...: from netCDF4 import Dataset
   ...: 
   ...: url = "https://thredds.aodn.org.au/thredds/dodsC/IMOS/SRS/ALTIMETRY/calibration_validation/geodeticGNSS/BASS/IMOS_SRS-Altimetry_WZ_20220307_B008-S602_FV02_22066-Geode
      ⋮ ticGNSS-imos080724-LowRate_END-20220504.nc"
   ...: 
   ...: crossover_date = cftime.DatetimeGregorian(1582, 10, 15)
   ...: 
   ...: with Dataset(url) as nc:
   ...:     time_var = nc["TIME"]
   ...:     times = cftime.num2date(time_var[:].compressed(), time_var.units, has_year_zero=True)
   ...: 
   ...: %time times < crossover_date
CPU times: user 1min 27s, sys: 9.9 ms, total: 1min 27s
Wall time: 1min 27s
Out[1]: array([False, False, False, ..., False, False, False], shape=(41494,))
```

[1] https://github.com/Unidata/cftime/blob/0019a7977ca17f0f350cae4f31e619e815c678d8/src/cftime/_cftime.pyx#L1417-L1446

ocefpaf:
I think that is similar to what I did in https://github.com/ioos/compliance-checker/pull/1212/files, but I'm still not 100% of my code there. I like your approach of just changing the calendar better.

Here in `cftime` I'm not sure how to optimize for that, in my example it is just 1 data point that requires transformation, so it is trivial to fix. I'm not sure how a generic solution for all cases would look like :-/

Maybe we can document the slowness and how that most of the time users will be comparing 1 data point vs a large array?

jswhit:
closed by PR #367

PR #367, opened 2025-07-10
speed up ```__richcmp__```

jswhit:
addresses issue #365.  Instead of converting both instances to ```proleptic_gregorian`` when calendars don't match, the LHS instance is converted to the calendar of the RHS instance.   Docstring for ``cftime.datetime``` is updated to suggest pre-converting instances to have same calendar with comparing an array to a scalar.


jswhit:
@ocefpaf could you have a look?

PR #371, opened 2025-09-16
MNT: unify static configuration files into `pyproject.toml`, update PEP 639 license metadata

neutrinoceros:
In preparation for a PR in the spirit of https://github.com/Unidata/netcdf4-python/pull/1427, modernize the project configuration around `pyproject.toml`


setup.py
```diff
@@ -2,14 +2,9 @@
 import sys
 import numpy
 
+from Cython.Build import cythonize
 from setuptools import Command, Extension, setup
 
-# https://github.com/Unidata/cftime/issues/34
```
neutrinoceros, line 7: note: direct invocations of `setup.py` as a scripts are deprecated in setuptools, so gh-34 should not be an issue now.

PR #372, opened 2025-09-21
WHL: build limited-api compliant wheels

neutrinoceros:
Based off #371, similar to Unidata/netcdf4-python#1427
I've included a commit to drop support for 3.8 and 3.9 (soon to be EOL), but it can easily be split out of this PR if desired.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,6 +42,9 @@ ABI3_TARGET_VERSION = "".join(str(_) for _ in sys.version_info[:2])
 ABI3_TARGET_HEX = hex(sys.hexversion & 0xFFFF00F0)
 
 if USE_PY_LIMITED_API:
+    DEFINE_MACROS  += [(("Py_LIMITED_API", ABI3_TARGET_HEX))]
+    
+if USE_PY_LIMITED_API:
     SETUP_OPTIONS = {"bdist_wheel": {"py_limited_api": f"cp{ABI3_TARGET_VERSION}"}}
 else:
     SETUP_OPTIONS = {}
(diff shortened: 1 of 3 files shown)


neutrinoceros:
`cp312-abi3` still errors (different symbol)
```
ImportError: dlopen(/private/var/folders/y6/nj790rtn62lfktb1sh__79hc0000gn/T/cibw-run-2th61c55/cp314-macosx_arm64/venv-test-arm64/lib/python3.14/site-packages/cftime/_cftime.abi3.so, 0x0002): symbol not found in flat namespace '__PyThreadState_UncheckedGet'
```
let's try `cp313-abi3`

ocefpaf:
> so that seems like a red herring.

BTW, I tried to build this locally for cp313 and ran the tests on cp314, everything works. I could not reproduce the error here :-/

neutrinoceros:
refreshing CI with a rebase just in case it works better this time (and because I forgot what the error looked like)

setup.py
```diff
@@ -60,6 +76,8 @@ def run(self):
     }
     DEFINE_MACROS += [('CYTHON_TRACE', '1'),
                      ('CYTHON_TRACE_NOGIL', '1')]
+    if USE_PY_LIMITED_API:
+        DEFINE_MACROS.append(("Py_LIMITED_API", ABI3_TARGET_HEX))
```
ocefpaf: I think I found the problem. These should probably be defined outside of this if-clause. I also believe that CYTHON_TRACE is not compatible with ABI3.

neutrinoceros: Nice. Feel free to push to this branch if you like. I'll be AFK until tomorrow

neutrinoceros:
Thanks a lot @ocefpaf. I've integrated your fix from #395 back here, so I expect CI to pass now.

PR #379, opened 2025-10-14
don't use cythonize in setup.py

jswhit:
@neutrinoceros hopefully this will help with PR #372

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,11 +74,7 @@ else:
     for e in ext_modules:
         e.cython_directives = {'language_level': "3"} 
         e.compiler_directives = COMPILER_DIRECTIVES
-
-if 'sdist' not in sys.argv[1:] and 'clean' not in sys.argv[1:] and '--version' not in sys.argv[1:]:
     # remove _cftime.c file if it exists, so cython will recompile _cftime.pyx.
-    # run for build *and* install (issue #263). Otherwise 'pip install' will
-    # not regenerate _cftime.c, even if the C lib supports the new features.
     if len(sys.argv) >= 2:
         if os.path.exists(CYTHON_CNAME):
             os.remove(CYTHON_CNAME)
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,6 @@ import os
 import sys
 import numpy
 
-from Cython.Build import cythonize
 from setuptools import Command, Extension, setup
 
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -68,16 +67,17 @@ def run(self):
...
-    )
+    ext_modules = [Extension('{}._{}'.format(NAME, NAME),
+                  sources=[os.path.relpath(CYTHON_FNAME, BASEDIR)],
+                  define_macros=DEFINE_MACROS,
+                  include_dirs=[numpy.get_include(),])]
+    for e in ext_modules:
+        e.cython_directives = {'language_level': "3"} 
+        e.compiler_directives = COMPILER_DIRECTIVES
+    # remove _cftime.c file if it exists, so cython will recompile _cftime.pyx.
+    if len(sys.argv) >= 2:
```
neutrinoceros: not sure what the intention for this check is. Seems to me there are many ways to end up executing this code with completely different commands so I'm a bit unconfortable with it. Can you explain ?

neutrinoceros (commented):
Thanks for the ping. I don't understand how this is supposed to work yet, and with a red CI I'm not yet convinced it should.

setup.py
jswhit: the idea is that anytime you run setup.py, you will want to recompile _cftime.pyx.  If there is a left-over _cftime.c file, then cython won't run, so this code removes it if it exists.

jswhit: From [the setuptools docs](https://setuptools.pypa.io/en/latest/userguide/ext_modules.html):

"When your [Cython](https://pypi.org/project/Cython) extension modules are declared using the [setuptools.Extension](https://setuptools.pypa.io/en/latest/userguide/ext_modules.html#setuptools.Extension) class, setuptools will detect at build time whether Cython is installed or not.

If Cython is present, then setuptools will use it to build the .pyx files. Otherwise, setuptools will try to find and compile the equivalent .c files (instead of .pyx). These files can be generated using the [cython command line tool](https://cython.readthedocs.io/en/stable/src/userguide/source_files_and_compilation.html)."

so you don't really need to use cythonize explicitly.

neutrinoceros: > the idea is that anytime you run setup.py, you will want to recompile _cftime.pyx

but why would this be captured specifically by the number of CLI arguments ? This is what I'm having a hard time wrapping my head around.

setup.py
```diff
@@ -68,16 +67,16 @@ def run(self):
...
-        compiler_directives=COMPILER_DIRECTIVES,
-        language_level=3,
-    )
+    ext_modules = [Extension('{}._{}'.format(NAME, NAME),
+                  sources=[os.path.relpath(CYTHON_FNAME, BASEDIR)],
+                  define_macros=DEFINE_MACROS,
+                  include_dirs=[numpy.get_include(),])]
+    for e in ext_modules:
+        e.cython_directives = {'language_level': "3"} 
+        e.compiler_directives = COMPILER_DIRECTIVES
```
ocefpaf, line 76: Looks like this part is being ignored. When I add `#cython: c_api_binop_methods=True` to the top of the pyx everything works as expected.

PR #388, opened 2026-01-08
add new cf calendar 'tai'

jswhit:
same as 'proleptic_gregorian', but only valid for dates after 1958-01-01.

exceptions are raised if dates before 1958 are requested  or a base date before 1958-01-01 is used.

Tests added by @ChrisBarker-NOAA

### DIFF
--- a/src/cftime/_cftime.pyx
+++ b/src/cftime/_cftime.pyx
@@ -30,7 +30,7 @@ _units = microsec_units+millisec_units+sec_units+min_units+hr_units+day_units
 # '366_day'=='all_leap','365_day'=='noleap')
 # see http://cfconventions.org/cf-conventions/cf-conventions#calendar
 # for definitions.
-_calendars = ['standard', 'gregorian', 'proleptic_gregorian','tai',
+_calendars = ['standard', 'gregorian', 'proleptic_gregorian', 'tai',
               'noleap', 'julian', 'all_leap', '365_day', '366_day', '360_day']
 _idealized_calendars= ['all_leap','noleap','366_day','365_day','360_day']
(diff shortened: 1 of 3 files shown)


jswhit:
@ChrisBarker-NOAA could you take a look?

src/cftime/_cftime.pyx
```diff
@@ -109,6 +109,9 @@ def _dateparse(timestr,calendar,has_year_zero=None):
     # parse the date string.
     year, month, day, hour, minute, second, microsecond, utc_offset =\
         _parse_date( isostring.strip() )
+    if calendar == 'tai':
+        if year != 1958 or month != 1 or day != 1 or minute != 0 or second != 0 or microsecond != 0:
```
ChrisBarker-NOAA: I may be missing what this code does, but I think we want to look for *before* 1958 -- it's OK to use a newer epoch, just not an older one.

But this check may be in the wrong place anyway -- if someone does negative "seconds since" then it could be wrong. (are negative numbers allowed at all? are they checked for?).

Could you compute the final dates, and then check they are all after 1958? (I guess that could get expensive).

If negative numbers are not allowed (checked for or not) then just checking the epoch should work fine.

jswhit: got it

ChrisBarker-NOAA (commented):
This looks good to me -- thanks Jeff! 

It could use some tests, though. *maybe* I'll find a bit of time to write a couple soon -- but don't wait for me.

ChrisBarker-NOAA:
Scratch that -- I see that the cftime.datetime class init doesn't take that keyword anyway.

PR #389, opened 2026-01-09
Add tests for tai calendar

ChrisBarker-NOAA:
I've added a (I think) complete set of tests for the tai calendar. I can't (quickly) figure out how to get coverage to work with Cython, so I'm only guessing, but I looked for 'tai' in the source, and tested all that I could find.

It only tests what's unique about tai, assuming the gregorian is already well tested.

I use "raw" pytest tests -- which I find easier to read and write, but will only work with pytest, but it seems you're using pytest, so all good.

But they could be ported to `unittest` if you'd like.

And a couple other tiny formatting updates ...

Thanks for getting on this!

### DIFF
--- a/src/cftime/_cftime.pyx
+++ b/src/cftime/_cftime.pyx
@@ -83,7 +83,7 @@ def _datesplit(timestr):
 
     return units.lower(), remainder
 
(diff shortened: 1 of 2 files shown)


jswhit:
looks good, thanks Chris!

PR #394, opened 2026-02-24
No cythonize take 2

ocefpaf:
Hopefully fix #379 and unlocks #372.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -72,8 +72,7 @@ else:
                   define_macros=DEFINE_MACROS,
                   include_dirs=[numpy.get_include(),])]
     for e in ext_modules:
-        e.cython_directives = {'language_level': "3"} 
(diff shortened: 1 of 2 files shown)


ocefpaf:
FYI this did not help #379. See my test where I tried both changes in https://github.com/ocefpaf/cftime/actions/runs/22368151216/job/64739567532

setup.py
neutrinoceros, line 16: I think most if not all of this comment could be removed if you're requiring Cython>=3.0 now

ocefpaf: If we are merging this I'll fix that, but this did not fix what we hoped and #372 still fails with this change. I'm not sure if there are other reasons for this change.

@jswhit what do you think? Should we close or finalize this one?

jswhit: if this doesn't fix #372, I don't think there is any reason to merge

PR #398, opened 2026-03-11
Use hashes in the GitHub Actions

ocefpaf:
This will make the GHA a bit more secure. One can even turn this on in the repo configuration or use a linter like zizmor to check the GHA. Note that I also cleaned up an unnecessary step that list the artifacts (they are shown and available for download in the GHA tab anyway) , and fix an if-clause that always evaluated to true.

Note that dependabot will update the GHA hash and the version in the comments, making this human friendly too.


.github/workflows/cibuildwheel.yml
```diff
@@ -110,14 +92,17 @@ jobs:
...
         pattern: pypi-artifacts*
         path: ${{ github.workspace }}/dist
         merge-multiple: true
 
-    - uses: pypa/gh-action-pypi-publish@release/v1
+    # Move to Trusted Publishing
+    - uses: pypa/gh-action-pypi-publish@ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e  # v1.13.0
       with:
         user: __token__
         password: ${{ secrets.PYPI_PASSWORD }}
```
ocefpaf, line 105: @jswhit the only security flaw found by zizmor that I cannot fix in this PR is here. It recommends the use of trusted publishers and this can be set on PyPI onlyby the owners of the package there. If you do, we can make the necessary changes here later.

jswhit: @ocefpaf I appear to be locked out of my pypi account. I've asked for a reset, just waiting for that to happen so I can do this.

jswhit: @ocefpaf I've regained access to pypi and added you as a maintainer to both cftime and netcdf4

ocefpaf: Note that I can't set trusted publishers as maintainers. Just admins can do that. It is not super important to do so, the setup here is safe enough, but is security is a concern, I do recommend to turn  trusted publishers on for both packages.

jswhit: OK, done for Unidata/cftime cibuildwheel.yml.  I didn't set an environment yet though.

ocefpaf: No, I need to know the name you used there and fix it here. We also need to remove the lines that uses the token. I can send a PR now.

PR #404, opened 2026-04-27
trusted publishers

ocefpaf:
Finishes #398


.github/workflows/cibuildwheel.yml
```diff
@@ -101,8 +101,9 @@ jobs:
     # Move to Trusted Publishing
     - uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b  # v1.14.0
       with:
-        user: __token__
-        password: ${{ secrets.PYPI_PASSWORD }}
         print_hash: true
+    environment:
+      name: release
```
ocefpaf, line 106: @jswhit this environment name, and the workflow file name itself, should match the one you used in the PyPI configuration.

PR #413, opened 2026-09-15
update for 1.6.6 release

jswhit:


jswhit:
python 3.14 wheel builds failing for Ubuntu (arm and x86_64)

jswhit:
Problem is manylinux-2014 has gcc 10.2, newest numpy requires 10.3

jswhit:
upgrading to manylinux_2_28 fixed the python 3.14 builds, but now python 3.15 is failing

PR #414, opened 2026-09-16
V1.6.6relprep remove twine

ocefpaf:
@jswhit I could nor commit on your branch, but just removing the unused `twine`, we will remove the cryptography build that is failing on Python 315t. With that said, we can do further cleanups. I'll tackle that on another PR after the release to avoid confusion.


jswhit:
thanks @ocefpaf !

Issue #412, opened 2026-09-11
New release


ocefpaf:
It would be nice to get a release with https://github.com/Unidata/cftime/pull/372 so we don't need to build wheels from 3.11 to 3.15 now that 3.15 is out.

jswhit:
PR #413  - python 3.15 builds are failing

ocefpaf:
I'll investigate this tomorrow. In a way, b/c of ABI3, those are test only. We won't ship wheels for py315. However, it is nice to understand what is going on before releasing.

jswhit:
all 1.6.6 release wheels built sucessfully except windows-11-arm - haven't had time to investigate why yet.

ocefpaf:
> all 1.6.6 release wheels built sucessfully except windows-11-arm - haven't had time to investigate why yet.

I'll check it tomorrow morning.

PR #415, opened 2026-09-17
Clean-ups

ocefpaf:
This PR started as an investigation to solve  https://github.com/Unidata/cftime/issues/412#issuecomment-5704917220, but I don't believe that windows-11-arm would ever pass on lower Python version. With ABI3 and in the light on the upcoming EoL of Python 3.10 we can clean the setup a little bit:

- Building Wheels only for  Python 3.11 (ABI3) and latest (for testing purposes);
- Dropped 3.10 from tests b/c it will be EoL next month;
- Dropped experimental CI, we needed that during the numpy 2.0 migration, now it is just redundant testing;
- Dropped x32 testing;
- Moved to a simpler setup without conda.

PS: There is no reason to use conda in the testing here. Different from netcdf4-python, where the netcdf-c, hdf5 from conda helps a lot, we can do an all pip setup for this repository. I'll change the CIs to reflect that here in a tomorrow.


neutrinoceros (commented):
as is, you're effectively skipping all free-threaded builds. I would recommend instead to omit `CIBW_BUILD` entirely so the whole array of supported versions is run.

pyproject.toml
```diff
@@ -107,19 +106,6 @@ ignore = [
     "test/*",
 ]
 
-[tool.cibuildwheel]
```
neutrinoceros, line 110: this table header should be preserved.

neutrinoceros:
> I don't believe that windows-11-arm would ever pass on lower Python version

Indeed, no version of numpy has ever supported Python 3.10 on windows arm, so it's no use doing it downstream.

pyproject.toml
ocefpaf: Just the header or all the config in there? I confess I'm not familiar with these config in the `pyproject.toml`, in all my projects I keep them in the cibuildwheel file.

neutrinoceros: I mean it depends what you wanted to do exactly, but I noticed that you only removed part of the table, and everything left is now meaningless without the header.

ocefpaf: Yeah, I was slowly moving the rest of the config there. But let's keep it as it was for now. There are already too many changes in this PR.

ocefpaf:
> as is, you're effectively skipping all free-threaded builds. 

I have 0 experience with the free-threaded builds. Should we do them here?

> I would recommend instead to omit `CIBW_BUILD` entirely so the whole array of supported versions is run.

My goal was to do to only oldest (forward compitable with ABO3) and latest (for future proofing) so I kept ``"cp311-* cp315-*"` there. As Python is getting more and more stable I don't see a reason to do them all. I'm fine removing that if you think it is worth to test them all. (We would need to re-do the PR skip from before.)

ocefpaf:
@neutrinoceros Let me know what you think. My idea is to reduce the maintenance burden and false positives, like the windows-arm one. I would like to remove more 😬 , like the x32 stuff. However, I think this is enough for now.

neutrinoceros:
Numpy stopped publishing wheels for win x32 a couple months back and as far as I could measure on astropy (using the same method as numpy folks) the downloads are really anecdotal.

pyproject.toml
```diff
@@ -27,7 +26,7 @@ dependencies = [
     "numpy>=2.3.0; platform_system == 'Windows' and platform_machine == 'ARM64'",
     "numpy>=1.21.2; platform_system != 'Windows' or platform_machine != 'ARM64'",
```
neutrinoceros: Supposedly, the only reason this dependency was forked was to prevent consumers from building unsupported versions from source, which in my opinion doesn't belong in cftime's metadata (otherwise there are *a lot* of similar forks missing here)
I also suggest bumping the minimum version to 1.23.2 as the oldest version with any cp311 wheels on PyPI
```suggestion
    "numpy>=1.23.2",
```

.github/workflows/cibuildwheel.yml
```diff
@@ -36,25 +36,11 @@ jobs:
...
-        echo "CIBW_TEST_SKIP=$CIBW_TEST_SKIP" >> $GITHUB_ENV
-        echo "Setting CIBW_TEST_SKIP=$CIBW_TEST_SKIP"
-
     - name: "Building ${{ matrix.os }} (${{ matrix.arch }}) wheels"
       uses: pypa/cibuildwheel@1828c10ab37f080699c7b81cea34097c684a7074  # v4.2.0
       env:
-        CIBW_SKIP: ${{ env.CIBW_SKIP }}
-        CIBW_TEST_SKIP: ${{ env.CIBW_TEST_SKIP }}
+        CIBW_SKIP: "*-musllinux*"
+        CIBW_BUILD: "cp311-* cp315-*"
```
neutrinoceros, line 43: I would actually suggest keeping as much as cibw's configuration as possible in `pyproject.toml`:
- a single source of truth is preferable to two
- static configuration is easier to maintain and reason with

neutrinoceros (commented):
a couple comments. I'm only looking at cibw configuration

pyproject.toml
```diff
@@ -24,10 +23,9 @@ classifiers = [
     'Topic :: Scientific/Engineering',
 ]
 dependencies = [
-    "numpy>=2.3.0; platform_system == 'Windows' and platform_machine == 'ARM64'",
-    "numpy>=1.21.2; platform_system != 'Windows' or platform_machine != 'ARM64'",
+    "numpy>=1.23.2",
 ]
-requires-python = ">=3.10"
+requires-python = ">=3.11"
```
neutrinoceros: btw setup.py (at least) has a `if sys.version_info >= (3, 11)` which could be simplified now

